PR #27, opened 2020-08-30
Pass version code and version name in build

LionZXY:


.github/workflows/release.yml
```diff
@@ -15,12 +15,12 @@ jobs:
         with:
           java-version: 1.8
       - name: Build release
```
rougsig, line 17: Build internal?

app/src/main/java/com/flipper/app/MainActivity.kt
```diff
@@ -16,7 +17,12 @@ class MainActivity : AppCompatActivity() {
     super.onCreate(savedInstanceState)
     val binding = ActivityMainBinding.inflate(layoutInflater)
     setContentView(binding.root)
-    initializeConductor(binding.root, savedInstanceState)
+    if (BuildConfig.DEBUG) {
+      binding.versionName.visibility = View.VISIBLE
```
rougsig, line 21: Can we use `binding.versionName.isVisible = true`?

rougsig (changes requested):

rougsig (approved):

PR #33, opened 2020-09-04
Add bottom navigation layout to home controller

rougsig:
device-2020-09-06-151447


app/src/main/java/com/flipper/app/home/ui/HomeView.kt
```diff
@@ -1,11 +1,11 @@
-package com.flipper.app.home
+package com.flipper.app.home.ui
 
+import com.flipper.app.home.ui.data.HomeTab
 import moxy.MvpView
 import moxy.viewstate.strategy.SingleStateStrategy
 import moxy.viewstate.strategy.StateStrategyType
 
 @StateStrategyType(SingleStateStrategy::class)
```
IlyaGulya, line 8: Moxy supports state strategy aliases.
For instance, istead of ` @StateStrategyType(SingleStateStrategy::class)` you can write `@SingleState`

IlyaGulya, line 8: Also, I'd suggest using `@AddToEndSingle` as default strategy for consistency across the project.

app/src/main/java/com/flipper/app/home/ui/data/HomeTab.kt
```diff
@@ -0,0 +1,26 @@
...
+            return when (itemId) {
+                Extensions.menuItemId -> Extensions
+                UserSpace.menuItemId -> UserSpace
+                Settings.menuItemId -> Settings
+                else -> {
+                    Timber.e("unknown item id in home tabs: itemId=${itemId}")
+                    null
+                }
+            }
+        }
```
IlyaGulya, line 24: How about rewriting this way?
In this case we will not need to maintain `createFromMenuItemId` method.
```
fun createFromMenuItemId(itemId: Int): HomeTab? {
    return values().firstOrNull { tab -> tab.menuItemId == itemId }.also { tab -> 
        if (tab == null) {
            Timber.e("unknown item id in home tabs: itemId=${itemId}")
        }
    }
}
```

app/src/main/java/com/flipper/app/stub/StubController.kt
```diff
@@ -0,0 +1,37 @@
...
+import com.bluelinelabs.conductor.RouterTransaction
+import com.flipper.app.databinding.ControllerStubBinding
+import com.flipper.core.view.BaseController
+import com.flipper.core.view.ViewInflater
+import moxy.presenter.InjectPresenter
+
+class StubController(args: Bundle) : BaseController<ControllerStubBinding>(args),
+    StubView {
+    @InjectPresenter
+    lateinit var presenter: StubPresenter
```
IlyaGulya, line 14: How about using `by moxyPresenter` instead?
```
val presenter by moxyPresenter {
  StubPresenter()
}
```
To use it you should add `implements MvpDelegateHolder` to `BaseController` and add `implementation "com.github.moxy-community:moxy-ktx:$moxyVersion"` dependency

core/src/main/java/com/flipper/core/view/BaseController.kt
```diff
@@ -4,32 +4,34 @@ import android.os.Bundle
...
 
     final override fun onCreateView(
         inflater: LayoutInflater,
         container: ViewGroup,
         savedViewState: Bundle?
     ): View {
+        if (!isCreated) {
+            isCreated = true
+            mvpDelegate.onCreate()
+        }
```
IlyaGulya, line 34: You should use `mvpDelegate.onCreate(savedViewState)` instead of checking for `isCreated` on your own.

core/src/main/java/com/flipper/core/view/BaseController.kt
```diff
@@ -56,10 +58,21 @@ abstract class BaseController<VB : ViewBinding> : Controller {
     }
 
     final override fun onDestroy() {
-        mvpDelegate.onDetach()
+        // If state is saved, it means controller rotation.
+        // Do not destroy presenter.
+        if (isStateSaved) return
+        mvpDelegate.onDestroy()
```
IlyaGulya, line 64: I think it would be better to use `!isBeingDestroyed` instead of `isStateSaved`

IlyaGulya (changes requested):

app/src/main/java/com/flipper/app/home/ui/HomeView.kt
LionZXY: Or write default state in gradle file

IlyaGulya: I think that manually assigning state strategy to each View interface is better because it's obvious which default strategy is used. If we will assign it as `apt` argument it will be less obvious.
That's why Moxy removed default strategy in 2.x version

core/src/main/java/com/flipper/core/view/BaseController.kt
rougsig: I'd prefer to keep isStateSaved, because i guess latter we will need to check if activity is finishing and for that `isBeingDestroyed` better name.

core/src/main/java/com/flipper/core/view/BaseController.kt
rougsig: See comment under isCreated field

core/src/main/java/com/flipper/core/view/BaseController.kt
IlyaGulya: No, you don't get it :smile: 
Conductor `Controller` already does all that stuff internally and exposes is as simple flag `isBeingDestroyed` which you can already use without needance to maintain it on your own.

rougsig: No. isBeingDestroyed is means rotation to.

LionZXY:
BottomNavigationView must fill all width

image

LionZXY (approved):

PR #35, opened 2020-09-04
Add detekt and ktlint

LionZXY:


.github/workflows/android.yml
```diff
@@ -18,6 +18,18 @@ jobs:
         run: ./gradlew assembleDebug
       - name: Unit-test
         run: ./gradlew test
+  static_analysis:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: ktlint
+        run: |
```
rougsig, line 26: Can we install ktlint and detekt as gradle plugins?
Prof: We can run that tasks manually before pushing to git

LionZXY: Now we can run manually also with ktlint/detekt bin

PR #62, opened 2021-09-08
Add echo screen

LionZXY:
**Background**

With newest firmware version you can use echo from/to flipper

**Changes**

- Add echo screen

**Test plan**

Open information screen
Type text and send

### DIFF
--- a/components/bridge/src/main/java/com/flipper/bridge/utils/Constants.kt
+++ b/components/bridge/src/main/java/com/flipper/bridge/utils/Constants.kt
@@ -17,6 +17,14 @@ object Constants {
         val SOFTWARE_VERSION: UUID = UUID.fromString("00002a28-0000-1000-8000-00805f9b34fb")
     }
 
+    // BLE serial service uuids: service uuid and characteristics uuids
+    object BLESerialService {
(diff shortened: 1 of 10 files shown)


components/bridge/src/main/java/com/flipper/bridge/utils/Constants.kt
```diff
@@ -17,6 +17,14 @@ object Constants {
...
+    object BLESerialService {
+        val SERVICE_UUID: UUID = UUID.fromString("8fe5b3d5-2e7f-4a98-2a48-7acc60fe0000")
+
+        val TX: UUID = UUID.fromString("19ed82ae-ed21-4c9d-4145-228e62fe0000")
+        val RX: UUID = UUID.fromString("19ed82ae-ed21-4c9d-4145-228e61fe0000")
+    }
+
     object BLE {
         const val RECONNECT_COUNT = 3
         const val RECONNECT_TIME = 100
```
rougsig, line 30: Sorry for add comment in unchanged line.

Its be more clear if constant contains unit of measurement time, like `RECONNECT_TIME_MS`.
if I need to change the constant, the first question will be: What is the unit of measurement of time here?

PR #65, opened 2021-09-17
Add pincode support

LionZXY:
**Background**

Add companion BLE feature

**Changes**

- Fix BLE
- Add companion BLE feature

**Test plan**


components/bridge/src/main/java/com/flipper/bridge/api/pair/FlipperPairApi.kt
```diff
@@ -1,12 +1,18 @@
...
 import android.content.Context
 import com.flipper.bridge.api.device.FlipperDeviceApi
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.TimeoutCancellationException
 import no.nordicsemi.android.ble.exception.BluetoothDisabledException
 
 interface FlipperPairApi {
+    fun getFlipperApi(
+        context: Context,
+        deviceId: String
```
rougsig, line 13: I think inline classes is better in that case.

https://kotlinlang.org/docs/inline-classes.html

components/bridge/src/main/java/com/flipper/bridge/impl/device/FlipperDeviceApiImpl.kt
```diff
@@ -4,7 +4,8 @@ import com.flipper.bridge.api.device.FlipperDeviceApi
 import com.flipper.bridge.impl.manager.FlipperBleManager
 
 class FlipperDeviceApiImpl(
-    private val bleManager: FlipperBleManager
+    private val bleManager: FlipperBleManager,
+    override val address: String
```
rougsig, line 8: And in that case 

https://kotlinlang.org/docs/inline-classes.html

components/bridge/src/main/java/com/flipper/bridge/impl/pair/FlipperPairApiImpl.kt
```diff
@@ -41,15 +54,19 @@ class FlipperPairApiImpl @Inject constructor(
         }
 
         val device = withTimeout(Constants.BLE.CONNECT_TIME_MS) {
-            scanner.findFlipperById(deviceId).first()
+            scanner.findFlipperById(flipperDeviceApi.address).first()
```
rougsig, line 57: findFlipperById or findFlipperByAddress?

components/pair/src/main/java/com/flipper/pair/api/PairComponentApiImpl.kt
```diff
@@ -19,10 +18,10 @@ class PairComponentApiImpl @Inject constructor(
         return preferences.getString(FlipperSharedPreferencesKey.DEVICE_ID, null) != null
     }
 
-    override fun getPairedDevice(): BLEDevice {
+    override fun getPairedDevice(): String {
```
rougsig, line 21: getPairedDeviceId?

components/pair/src/main/java/com/flipper/pair/find/service/PairDeviceViewModel.kt
```diff
@@ -0,0 +1,49 @@
...
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.collect
+import kotlinx.coroutines.launch
+import no.nordicsemi.android.ble.ktx.state.ConnectionState
+import no.nordicsemi.android.ble.ktx.stateAsFlow
+
+class PairDeviceViewModel(application: Application) : AndroidViewModel(application) {
+    private val context = application
+    private val _state = MutableStateFlow<ConnectionState?>(null)
```
rougsig, line 17: Check it, i think one stateFlow with one big state is more useful, you can check sample here https://airbnb.io/mavericks/#/

components/pair/src/main/java/com/flipper/pair/api/PairComponentApiImpl.kt
rougsig, line 21: Yeap i think will be better here, less way to misstype.

https://kotlinlang.org/docs/inline-classes.html

components/bridge/src/main/java/com/flipper/bridge/impl/device/FlipperDeviceApiImpl.kt
LionZXY: Not sure, because we want also provide public api in FlipperDeviceApi

PR #103, opened 2021-11-04
Add screenstreaming

LionZXY:
**Background**

We already have api for screenstreaming

**Changes**

So we just add screen streaming over BLE

**Test plan**

- Open settings tab


components/screenstreaming/impl/src/main/java/com/flipperdevices/screenstreaming/impl/viewmodel/ScreenStreamFrameDecoder.kt
```diff
@@ -0,0 +1,46 @@
...
+            for (y in 0 until SCREEN_HEIGHT) {
+                val color = if (bytes.isPixelSet(x, y)) Color.BLACK else BACKGROUND_COLOR
+                screen[x, y] = color
+            }
+        }
+        return screen
+    }
+
+    private fun ByteArray.isPixelSet(x: Int, y: Int): Boolean {
+        var index = (y / Byte.SIZE_BITS) * SCREEN_WIDTH
```
rougsig, line 41: Why not just `(y / Byte.SIZE_BITS) * SCREEN_WIDTH + x`?

components/screenstreaming/impl/src/main/java/com/flipperdevices/screenstreaming/impl/viewmodel/ScreenStreamingViewModel.kt
```diff
@@ -0,0 +1,104 @@
...
+    fun onStartStreaming() {
+        serviceApi?.requestApi?.request(
+            main {
+                guiStartScreenStreamRequest = startScreenStreamRequest {}
+            }
+        )?.onEach {
+            onStreamFrameReceived(it.guiScreenStreamFrame)
+        }?.launchIn(viewModelScope)
+
+        serviceApi?.requestApi?.notificationFlow()?.onEach {
```
rougsig, line 53: I think after `onPauseStreaming` that `onEach` is active. It is?

PR #104, opened 2021-11-08
Add buffer overflow support

LionZXY:
**Background**

Now we can't send many pocket while device not ready

**Changes**

- Read BLE charactertics 
- Send request only if buffer available

**Test plan**

Press on "Send 1000 pocket"


components/bridge/impl/src/main/java/com/flipperdevices/bridge/impl/manager/PeripheralResponseReader.kt
```diff
@@ -1,32 +1,33 @@
...
 
-@ObsoleteCoroutinesApi
 @Suppress("BlockingMethodInNonBlockingContext")
 class PeripheralResponseReader(
     private val scope: CoroutineScope
 ) : LogTagProvider {
     override val TAG = "PeripheralResponseReader"
     private val byteInputStream = ByteEndlessInputStream()
     private val responses = MutableSharedFlow<Flipper.Main>()
+    private val responseReaderDispatcher = newSingleThreadExecutor(TAG)
```
rougsig, line 25: Why not IO dispatcher?

components/bridge/impl/src/main/java/com/flipperdevices/bridge/impl/manager/overflow/FlipperRequestComparator.kt
```diff
@@ -0,0 +1,9 @@
+package com.flipperdevices.bridge.impl.manager.overflow
+
+import com.flipperdevices.bridge.api.model.FlipperRequest
+
+class FlipperRequestComparator : Comparator<FlipperRequest> {
+    override fun compare(first: FlipperRequest, second: FlipperRequest): Int {
+        return first.priority.ordinal - second.priority.ordinal
```
rougsig, line 7: It's very hacky, i think you should add comment to priority enum like: 'Do not reorder enum declaration list, in reason ... ...'. 

Or make code more clear with adding level field:
```
// Lower level is more priority task
enum Priority(val level: Int) {
  Main(level = 1),
  BG(level = 2),
  ETC(level = 3),
}
```

components/bridge/impl/src/main/java/com/flipperdevices/bridge/impl/manager/overflow/FlipperRequestStorageImpl.kt
```diff
@@ -0,0 +1,22 @@
...
+
+import com.flipperdevices.bridge.api.model.FlipperRequest
+import java.util.concurrent.PriorityBlockingQueue
+import java.util.concurrent.TimeUnit
+
+private const val QUEUE_INITIAL_CAPACITY = 11
+
+class FlipperRequestStorageImpl : FlipperRequestStorage {
+    override val TAG = "FlipperRequestStorage"
+    private val queue = PriorityBlockingQueue(QUEUE_INITIAL_CAPACITY, FlipperRequestComparator())
```
rougsig, line 11: PriorityBlockingQueue is from java.concurrent lib. With coroutines you should not to use that package. Usage can create strange bugs. 

Check thouse issues:
https://github.com/Kotlin/kotlinx.coroutines/issues/238
https://github.com/Kotlin/kotlinx.coroutines/issues/261

PR #114, opened 2021-11-18
use dependencyResolutionManagement block

dsvoronin:
**Background**

allprojects/subprojects blocks should be avoided, because it adds additional complexity for build configuration, that could be avoided by using convention plugins and special contructs as `dependencyResolutionManagement`, also it blocks configureOnDemand and incoming ['project-isolation'](https://gradle.github.io/configuration-cache/#constraints_for_build_scripts_and_plugins) feature

**Changes**

Use new centralized block for dependency resolution

**Test plan**

dependency resolution trigger for all configuration will show a problem with repository configuration
simple current PR build will be enough

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -6,14 +6,6 @@ plugins {
     id("org.jlleitschuh.gradle.ktlint") version GradlePlugins.KTLINT
 }
 
-allprojects {
-    repositories {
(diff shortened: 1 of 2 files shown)


LionZXY (approved):
LGTM. Thank you!

PR #200, opened 2022-04-06
Fix trash category

LionZXY:
**Background**

Now we can restore key from key screen, but can't do it for "All" keys

**Changes**

- Add restore all button
- Add delete all button


LionZXY:
**Background**

Now we can restore key from key screen, but can't do it for "All" keys

**Changes**

- Add restore all button
- Add delete all button

PR #212, opened 2022-04-28
Fix possibility cancel editing keys

Programistich:
**Background**

Now we can cancel editing key

**Changes**

* Add condition equals old/new key after cancel editing key

**Test plan**

* Choose key
* Select Edit
* Do something changes
* Click 'Cancel'
* Look for key information didn't change

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 # 1.1 - In progress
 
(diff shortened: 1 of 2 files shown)


LionZXY (dismissed):
Good work!

LionZXY (approved):

PR #214, opened 2022-05-01
Protobuf library

Programistich:
**Background**

Our library for protobuf

**Changes**

* Delete git submodule
* Implementation library
* Change kotlin and java-lite to kotlin-lite
* Rename protobuf module

**Test plan**

 * Build and start app


components/bridge/api/build.gradle.kts
```diff
@@ -14,4 +14,7 @@ dependencies {
     implementation(libs.ble.scan)
 
     implementation(libs.appcompat)
+
+    // Protobuf jvm
+    implementation(libs.protobuf.jvm)
```
LionZXY, line 19: Why not locate 
`implementation(libs.protobuf.jvm)` near with `implementation(projects.components.bridge.pbutils)`?

PR #215, opened 2022-05-02
Compose Button write and emulate

Programistich:
**Background**

* Compose Button write and emulate

**Changes**

* String resources "write" and "emulate"
* Icon resources "write" and "emulate"
* Common Compose View for Action Flipper Button
* Compose View for Write and Emulate

**Test plan**

* Start Application

image


components/keyscreen/impl/src/main/java/com/flipperdevices/keyscreen/impl/composable/actions/ComposableActionFlipper.kt
```diff
@@ -0,0 +1,71 @@
...
+    @ColorRes descriptionColorId: Int = com.flipperdevices.core.ui.R.color.white_100,
+    @ColorRes tintId: Int = com.flipperdevices.core.ui.R.color.white_100,
+    onClick: () -> Unit,
+) {
+    val descriptionText = stringResource(descriptionId)
+
+    Box(
+        modifier = Modifier
+            .clip(shape = RoundedCornerShape(30.dp))
+            .weight(1f)
```
LionZXY, line 46: Please, move row-specific modifier arguments in Row and then pass modifier in function

components/keyscreen/impl/src/main/java/com/flipperdevices/keyscreen/impl/composable/actions/ComposableActionFlipper.kt
```diff
@@ -0,0 +1,71 @@
...
+            .background(colorResource(id = DesignSystem.color.accent_secondary))
+            .clickable { onClick() }
```
LionZXY, line 48: On all clickable element you should add android ripple effect

components/keyscreen/impl/src/main/java/com/flipperdevices/keyscreen/impl/composable/ComposableKeyParsed.kt
```diff
@@ -43,6 +51,16 @@ fun ComposableKeyParsed(
             keyScreenState.favoriteState,
             viewModel::setFavorite
         )
+        Row(
+            modifier = Modifier
+                .fillMaxWidth()
+                .padding(start = 24.dp, end = 24.dp, bottom = 24.dp),
+        ) {
+            ComposableWrite(onClick = {})
+            Spacer(modifier = Modifier.width(36.dp))
```
LionZXY, line 60: Why not use padding passing in functions?

components/keyscreen/impl/src/main/java/com/flipperdevices/keyscreen/impl/composable/actions/ComposableActionFlipper.kt
```diff
@@ -0,0 +1,71 @@
...
+        Column(
+            modifier = Modifier
+                .fillMaxWidth()
+                .padding(vertical = 4.dp),
+            verticalArrangement = Arrangement.Center,
+            horizontalAlignment = Alignment.CenterHorizontally,
+        ) {
+            Icon(
+                modifier = Modifier.size(size = 36.dp),
+                painter = painterResourceByKey(iconId),
```
LionZXY, line 59: Why you use `painterResourceByKey` instead of `painterResource` here?

components/keyscreen/impl/src/main/java/com/flipperdevices/keyscreen/impl/composable/actions/ComposableActionFlipper.kt
```diff
@@ -0,0 +1,71 @@
...
+                contentDescription = descriptionText,
+                tint = colorResource(tintId)
+            )
+            Text(
+                text = descriptionText,
+                fontWeight = FontWeight.Bold,
```
LionZXY, line 65: Please, use `W500`/`W400`/`W600` instead of `Bold` and etc, because designer use number for weight

components/keyscreen/impl/src/main/java/com/flipperdevices/keyscreen/impl/composable/actions/ComposableWrite.kt
```diff
@@ -0,0 +1,28 @@
+package com.flipperdevices.keyscreen.impl.composable.actions
+
+import androidx.compose.runtime.Composable
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.RowScope
+import androidx.compose.ui.tooling.preview.Preview
+import com.flipperdevices.keyscreen.impl.R
+
+@Composable
+fun RowScope.ComposableWrite(onClick: () -> Unit) {
```
LionZXY, line 10: NIT: You can add this function in `ComposableActionFlipper.kt` file

components/keyscreen/impl/src/main/java/com/flipperdevices/keyscreen/impl/composable/actions/ComposableActionFlipper.kt
Programistich: What kind of modifiers should be moved and where? Top level up in a row?

components/keyscreen/impl/src/main/java/com/flipperdevices/keyscreen/impl/composable/ComposableKeyParsed.kt
Programistich: in what functions

components/keyscreen/impl/src/main/java/com/flipperdevices/keyscreen/impl/composable/actions/ComposableActionFlipper.kt
LionZXY: Row modifier (like `weight`) can be moved in row. You should also ask yourself, "Can I use this component elsewhere? What is stopping me? How can I fix it?

components/keyscreen/impl/src/main/java/com/flipperdevices/keyscreen/impl/composable/ComposableKeyParsed.kt
LionZXY: ComposableWrite and e.t.c.
I think it's better to avoid unnecessary elements

PR #222, opened 2022-05-16
Minor grammar corrections in the settings screen

Astrrra:
**Background**

The settings screen looked a bit clunky because of the wording of some labels there

**Changes**

Re-word a few strings so they sound a bit better

**Test plan**

Build the app, open the settings screen, look at the text there


LionZXY (changes requested):
Please, fix compile

PR #223, opened 2022-05-18
Try use Jetpack Compose Navigation in "Options"

Programistich:
**Background**

* Try to use Compose Navigation in screen "Options"

**Changes**

* Cicerone screen change to Composable screen
* Create NavGraph
* realization BackPressListener

**Test plan**

* Start application, go to "Options" and navigate to different screens


components/screenstreaming/impl/src/main/java/com/flipperdevices/screenstreaming/impl/composable/ComposableStreamingScreen.kt
```diff
@@ -0,0 +1,45 @@
...
+import com.flipperdevices.core.ktx.android.observeAsState
+import com.flipperdevices.screenstreaming.impl.model.StreamingState
+import com.flipperdevices.screenstreaming.impl.viewmodel.ScreenStreamingViewModel
+
+@Composable
+fun ComposableStreamingScreen(
+    screenStreamingViewModel: ScreenStreamingViewModel = viewModel()
+) {
+    val lifecycleState = LocalLifecycleOwner.current.lifecycle.observeAsState()
+    if (lifecycleState.value == Lifecycle.Event.ON_PAUSE) {
```
LionZXY, line 16: Here maybe better rework to delegate
```
val lifecycleState by LocalLifecycleOwner.current.lifecycle.observeAsState()
```

I am not sure that current code work correctly when state changing

components/screenstreaming/impl/src/main/java/com/flipperdevices/screenstreaming/impl/composable/ComposableStreamingScreen.kt
```diff
@@ -0,0 +1,45 @@
...
+        screenStreamingViewModel.getStreamingState().compareAndSet(
```
LionZXY, line 17: Please, move all modification of state to view model and send to compose only read-only state

components/screenstreaming/impl/src/main/java/com/flipperdevices/screenstreaming/impl/composable/ComposableStreamingScreen.kt
```diff
@@ -0,0 +1,45 @@
...
+    ComposableScreen(
+        screenStreamingViewModel,
+        onPressButton = { button ->
+            screenStreamingViewModel.onPressButton(button)
+        },
+        onLongPressButton = { button ->
+            screenStreamingViewModel.onLongPressButton(button)
+        },
+        onScreenStreamingSwitch = { state ->
+            if (state == StreamingState.ENABLED) {
```
LionZXY, line 32: Maybe change `if` to `when`? Because in future we can add one more value in enum and in case with `when` we will have compilation error in this case

components/screenstreaming/impl/src/main/java/com/flipperdevices/screenstreaming/impl/fragment/ScreenStreamingFragment.kt
```diff
@@ -7,6 +7,7 @@ import com.flipperdevices.screenstreaming.impl.composable.ComposableScreen
 import com.flipperdevices.screenstreaming.impl.model.StreamingState
 import com.flipperdevices.screenstreaming.impl.viewmodel.ScreenStreamingViewModel
 
+@Deprecated(message = "Replace with compose navigation")
 class ScreenStreamingFragment : ComposeFragment() {
```
LionZXY, line 11: Why not delete it?

components/screenstreaming/impl/src/main/java/com/flipperdevices/screenstreaming/impl/composable/ComposableStreamingScreen.kt
Programistich: Maybe  
screenStreamingViewModel.getStreamingState().compareAndSet(
expect = StreamingState.ENABLED,
update = StreamingState.DISABLED
)

to take out in separate method? it repeats here and in onPause

LionZXY (approved):
Very good job! Thank you!

PR #231, opened 2022-06-08
Add image on Flipper updating

LionZXY:
**Background**

Now we not pause flipper when updating

**Changes**

- Upload image with reverse streaming 

**Test plan**

Try update flipper


components/updater/impl/src/main/java/com/flipperdevices/updater/impl/UpdaterTask.kt
```diff
@@ -90,6 +93,7 @@ class UpdaterTask(
         onStateUpdate: suspend (UpdatingState) -> Unit
     ) = FlipperStorageProvider.useTemporaryFolder(context) { tempFolder ->
         info { "Start update with folder: ${tempFolder.absolutePath}" }
+
```
Programistich, line 96: New line

LionZXY: This is a feature. Do you want remove it?

Programistich (approved):

PR #239, opened 2022-06-14
Add dialog on flipper discharged

LionZXY:
**Background**

For now we allow update flipper even if flipper is discharded

**Changes**

- Show dialog that suggest charge flipper

telegram-cloud-photo-size-2-5280524761740459115-y


Programistich (approved):
Good work

PR #254, opened 2022-06-27
Compose theme

Programistich:
**Background**

Migrate from xml theme to compose

**Changes**

* Pallet
* Typography
* CustomTheme with provides Locals

**Test plan**

..and how we can test it


LionZXY (changes requested):

components/core/ui/fragment/src/main/java/com/flipperdevices/core/ui/fragment/theme/FlipperTypography.kt
```diff
@@ -0,0 +1,40 @@
...
+    val bodyR14: TextStyle,
+    val bodyM14: TextStyle,
+    val bodyR16: TextStyle,
+
+    val buttonB16: TextStyle,
+    val buttonM16: TextStyle,
+    val buttonB14: TextStyle
+)
+
+internal val fontFamily = FontFamily(
```
LionZXY, line 27: I remember that you don't like to use human names for font weights. 
I agree with you, maybe the best way - use numeric weight (like W500/100 e.t.c) for defining the font family.  What you think about this?

Programistich: Agree with u

PR #260, opened 2022-06-29
Compose theme - new module `theme`

Programistich:
**Background**

* Re-split fragment on new module for future compose only

**Changes**

* New module with theme components(pallet, typography)
* Fonts move to res module

**Test plan**

* Build app


LionZXY:
Good job, thank you!

PR #261, opened 2022-06-29
Update card not shown when flipper not connected

LionZXY:
**Background**

For now we not show update card

**Changes**

- Split `NotReady` to `NotConnected` and `ConnectingInProgress`

**Test plan**

Try forget flipper and see state on `NotConnected`


Programistich:
Good work

PR #264, opened 2022-06-30
Сompose theme - moduls `bottom bar` and `connection`

Programistich:
**Background**

* Move to compose theme

**Changes**

* Replace colors
* Replace typography
* Move drawable and animations
* Connection API return ConnectionState, and in BottomBar from ConnectionState  to TabState (with Composable scope )
**Test plan**

* Install and check bottom bar states


components/bottombar/impl/src/main/java/com/flipperdevices/bottombar/impl/main/compose/BottomTabMapper.kt
```diff
@@ -1,104 +0,0 @@
...
-import com.flipperdevices.bottombar.impl.model.FlipperBottomTab
-import com.flipperdevices.bottombar.model.TabState
-import com.flipperdevices.connection.api.api.ConnectionApi
-import com.flipperdevices.connection.api.model.ConnectionStatusState
-import com.flipperdevices.core.ktx.jre.roundPercentToString
-import com.flipperdevices.core.ui.res.R as DesignSystem
-import com.flipperdevices.core.ui.theme.LocalPallet
-
-@Composable
-fun getTabStateFromFlipperBottomTab(
```
LionZXY, line 14: Why you move this method?

components/bottombar/api/src/main/java/com/flipperdevices/bottombar/model/TabState.kt
```diff
@@ -1,28 +1,28 @@
 package com.flipperdevices.bottombar.model
 
-import androidx.annotation.ColorRes
 import androidx.annotation.DrawableRes
 import androidx.annotation.RawRes
+import androidx.compose.ui.graphics.Color
 
 private const val MAGIC_NUMBER = 31
 
 sealed class TabState private constructor(
```
LionZXY, line 9: Please, write unit test on equals/hashcode method for this class

PR #271, opened 2022-07-13
Switch theme

Programistich:
**Background**

Switch theme in options - system, dark, light

**Changes**

* Protobuf enum - SelectedTheme(dark, light, system)
* DropDownMenu with values from enum
* Remove parameter isLight
* ThemeViewModel get SelectedTheme
* Launched Effect on change theme

**Test plan**

..and how we can test it

### DIFF
--- a/components/core/ui/theme/src/main/kotlin/com/flipperdevices/core/ui/theme/FlipperTheme.kt
+++ b/components/core/ui/theme/src/main/kotlin/com/flipperdevices/core/ui/theme/FlipperTheme.kt
@@ -9,7 +9,7 @@ import androidx.compose.material.Shapes
 import androidx.compose.material.contentColorFor
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.CompositionLocalProvider
-import androidx.compose.runtime.DisposableEffect
+import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.compositionLocalOf
 import androidx.compose.runtime.getValue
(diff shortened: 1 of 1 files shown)


components/core/ui/theme/src/main/kotlin/com/flipperdevices/core/ui/theme/FlipperTheme.kt
```diff
@@ -1,31 +1,45 @@
...
-    isLight: Boolean,
+    themeViewModel: ThemeViewModel = viewModel(),
     content: @Composable () -> Unit
 ) {
-    val pallet = if (isLight) lightPallet else darkPallet
+    val theme by themeViewModel.appTheme().collectAsState()
+    val isLight = isLight(theme)
+    changeXMLColor(theme)
+
+    val pallet = (if (isLight) lightPallet else darkPallet).switch()
```
LionZXY, line 42: Why you use `isLight` variable?

components/core/ui/theme/src/main/kotlin/com/flipperdevices/core/ui/theme/FlipperTheme.kt
```diff
@@ -42,6 +56,24 @@ fun FlipperTheme(
...
+}
+
+private fun changeXMLColor(theme: SelectedTheme) {
+    val systemThemeId = when (theme) {
+        SelectedTheme.DARK -> AppCompatDelegate.MODE_NIGHT_YES
+        SelectedTheme.LIGHT -> AppCompatDelegate.MODE_NIGHT_NO
+        else -> AppCompatDelegate.MODE_NIGHT_FOLLOW_SYSTEM
+    }
+    AppCompatDelegate.setDefaultNightMode(systemThemeId)
+}
```
LionZXY, line 75: Move to disposable effect

Issue #234, opened 2022-06-09
updater: Local firmware build detected as "Release", unlike qFlipper


digitalcircuit:
**Describe the bug**
The Flipper Android app assumes `FirmwareChannel.RELEASE` if it is not able to parse the firmware version and incorrectly offers to install an update.  This contrasts with qFlipper, which assumes `CUSTOM` if it fails to parse the version.

**To Reproduce**
1. Build the Flipper firmware locally in a branch **not** named `dev`, `rc`, or `release`
2. Install your local firmware build
3. Connect to the Flipper Android app
4. Observe the `Firmware Version` and `Firmware Update` sections

**Expected behavior**
The Flipper Android app correctly detects the use of a local firmware build and does not attempt to offer an update.

However, the Flipper Android app should still provide a way to change the `Update Channel` back to `Release`, `Release Candidate`, or `Development` to provide a way to migrate away from a local firmware build.

**Logs**
*Unsure of which logs would be relevant - a full `adb logcat`…?*

**Target**
Pixel 4 XL running Android 12 (SP2A.220505.002)

**Additional context**
The Flipper Android app [builds the firmware version from a string in `buildFirmwareVersionFromString()` from `FirmwareVersionBuildHelper.kt`](https://github.com/flipperdevices/Flipper-Android-App/blob/d20ef4bb82ea71486b7b3727712e05eb8fb9cc2a/components/updater/impl/src/main/java/com/flipperdevices/updater/impl/utils/FirmwareVersionBuildHelper.kt#L14-L45 ).  This falls through to `FirmwareChannel.RELEASE`.

qFlipper [builds the firmware version from a string in `branchToChannelName()` from `deviceinfohelper.cpp`](https://github.com/flipperdevices/qFlipper/blob/73c2ac6ac119cfaf5cabbc5c24f7cfc1a3f4b582/backend/flipperzero/helper/deviceinfohelper.cpp#L262-L283 ).  This falls through to `CUSTOM`.

To fix this in the Android app in a way that allows reverting to a release/RC/dev build, this may involve adding a new firmware channel that can't be manually selected, but the app selects it when it can't parse the version.

digitalcircuit:
It's more of a visual bug - the app claims there's an update available (big, green `Update` button), when instead it should act as if switching update channel, offering to replace the firmware (orange `Install` button).

## qFlipper behavior
[...]

PR #281, opened 2022-07-18
Support unknown firmware

Programistich:
**Background**

Support unknown firmware https://github.com/flipperdevices/Flipper-Android-App/issues/234

**Changes**

* New value for enum FirmwareChanel
* Update flag 'always update' if we connected with unknown firmware 
* Update parser version string to FirmwareVersion (like qFlipper https://github.com/flipperdevices/qFlipper/blob/7aeb2b595da4eabdfe16e270834777c2c1033aa7/backend/flipperzero/helper/deviceinfohelper.cpp#L262) + test

**Test plan**

Connect application to flipper with unknown firmware

### DIFF
--- a/components/info/shared/src/main/java/com/flipperdevices/info/shared/ComposableFirmwareVersion.kt
+++ b/components/info/shared/src/main/java/com/flipperdevices/info/shared/ComposableFirmwareVersion.kt
@@ -14,7 +14,7 @@ fun getNameByChannel(channel: FirmwareChannel): Int {
         FirmwareChannel.DEV -> R.string.info_device_firmware_version_dev
         FirmwareChannel.RELEASE -> R.string.info_device_firmware_version_release
         FirmwareChannel.RELEASE_CANDIDATE -> R.string.info_device_firmware_version_rc
-        FirmwareChannel.UNKNOWN -> R.string.info_device_firmware_version_unknown
+        FirmwareChannel.UNKNOWN -> R.string.info_device_firmware_version_unknown_empty
     }
 }
 
@@ -24,7 +24,7 @@ fun getFullNameByChannel(channel: FirmwareChannel): Int {
         FirmwareChannel.DEV -> R.string.info_device_firmware_version_dev_full
         FirmwareChannel.RELEASE -> R.string.info_device_firmware_version_release_full
         FirmwareChannel.RELEASE_CANDIDATE -> R.string.info_device_firmware_version_rc_full
-        FirmwareChannel.UNKNOWN -> R.string.info_device_firmware_version_unknown_full
+        FirmwareChannel.UNKNOWN -> R.string.info_device_firmware_version_unknown_empty
     }
 }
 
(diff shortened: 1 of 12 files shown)


components/info/shared/src/main/java/com/flipperdevices/info/shared/ComposableFirmwareVersion.kt
```diff
@@ -32,12 +34,16 @@ fun getDescriptionByChannel(channel: FirmwareChannel): Int {
...
     }
 }
 
 @Composable
 fun getTextByVersion(version: FirmwareVersion): String {
-    return "${stringResource(getNameByChannel(version.channel))} ${version.version}"
+    return when (version.channel) {
+        FirmwareChannel.UNKNOWN -> version.version
+        else -> "${stringResource(getNameByChannel(version.channel))} ${version.version}"
+    }
```
LionZXY, line 46: It doesn't seem safe to me when we handle the `Unknown` case separately. Is it possible not to make such a branch here? Maybe `getNameByChannel(version.channel)` will return an empty string?

components/updater/card/src/main/java/com/flipperdevices/updater/card/viewmodel/UpdateCardViewModel.kt
```diff
@@ -215,5 +216,6 @@ private fun FirmwareChannel?.toSelectedChannel(): SelectedChannel = when (this)
     FirmwareChannel.RELEASE -> SelectedChannel.RELEASE
     FirmwareChannel.RELEASE_CANDIDATE -> SelectedChannel.RELEASE_CANDIDATE
     FirmwareChannel.DEV -> SelectedChannel.DEV
+    FirmwareChannel.UNKNOWN -> SelectedChannel.UNRECOGNIZED
```
LionZXY, line 219: Let's throw error here, what you think? Or just print `error` in log

components/updater/impl/src/main/java/com/flipperdevices/updater/impl/utils/FirmwareVersionBuildHelper.kt
```diff
@@ -26,7 +28,7 @@ object FirmwareVersionBuildHelper {
             return FirmwareVersion(FirmwareChannel.DEV, hash, date)
         }
 
-        if (typeVersion.contains(DEVICE_VERSION_TYPE_RC)) {
+        if (DEVICE_VERSION_TYPE_RC_REGEX.toRegex() matches typeVersion) {
```
LionZXY, line 31: It seems we need to prepare the data here. For example, `1.0.0-rc ` will fail because of an extra space at the end.

components/updater/impl/src/test/kotlin/com/flipperdevices/updater/impl/FirmwareVersionBuildHelperTest.kt
```diff
@@ -0,0 +1,70 @@
...
+import com.flipperdevices.updater.impl.utils.FirmwareVersionBuildHelper
+import com.flipperdevices.updater.model.FirmwareChannel
+import com.flipperdevices.updater.model.FirmwareVersion
+import org.junit.Assert
+import org.junit.Test
+
+class FirmwareVersionBuildHelperTest {
+
+    @Test
+    fun `Correct dev firmware`() {
```
LionZXY, line 12: NIT: You can use _JUnit Parameterized Test_

components/updater/impl/src/test/kotlin/com/flipperdevices/updater/impl/FirmwareVersionBuildHelperTest.kt
```diff
@@ -0,0 +1,70 @@
...
+            channel = FirmwareChannel.UNKNOWN,
+            version = "name/2204_bt_forget_devices02d45365",
+            buildDate = "13-07-2022"
+        )
+        Assert.assertEquals(parseFirmware, actualFirmware)
+    }
+
+    @Test
+    fun `Uncorrected device version part count`() {
+        val firmwareVersion = "3e5d499b12345671363 13-07-2022"
```
LionZXY, line 65: NIT: You can use the top versions of the firmware for the test
image

components/updater/impl/src/main/java/com/flipperdevices/updater/impl/utils/FirmwareVersionBuildHelper.kt
Programistich: Where we get word with whitespace, if we filter "firmwareVersion" stream by not empty?

PR #287, opened 2022-07-27
NFC Editor perfomance optimization

LionZXY:
**Background**

Now we have huge lags with NFC editor. Also nfc editor not work correctly with text

**Changes**

- Optimize compose layout
- Add `TextUpdaterHelper`

**Test plan**

Try open sample on phone and try edit some cards

### DIFF
--- a/components/nfceditor/impl/src/main/java/com/flipperdevices/nfceditor/impl/composable/ComposableNfcEditor.kt
+++ b/components/nfceditor/impl/src/main/java/com/flipperdevices/nfceditor/impl/composable/ComposableNfcEditor.kt
@@ -10,8 +10,10 @@ import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.text.TextStyle
 import androidx.lifecycle.viewmodel.compose.viewModel
+import com.flipperdevices.core.ktx.jre.length
 import com.flipperdevices.core.ui.theme.LocalPallet
 import com.flipperdevices.core.ui.theme.LocalTypography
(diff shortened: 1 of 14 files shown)


components/nfceditor/impl/src/main/java/com/flipperdevices/nfceditor/impl/composable/ComposableNfcEditor.kt
```diff
@@ -25,41 +28,41 @@ fun ComposableNfcEditor(nfcEditorViewModel: NfcEditorViewModel = viewModel()) {
...
-                    onValueChanged = { location, textValue ->
-                        nfcEditorViewModel.onChangeSelection(
-                            location,
-                            textValue.selection.start
-                        )
-                    }
-                )
-            }
+private fun ComposableNfcEditor(
+    nfcEditorViewModel: NfcEditorViewModel,
```
Programistich, line 52: NIT: Maybe we can not provide viewmodel, only needed method/field?

Programistich (approved):

components/nfceditor/impl/src/main/java/com/flipperdevices/nfceditor/impl/composable/ComposableNfcEditor.kt
LionZXY: As I understand it is better to avoid reading parameters in places where performance is important to reduce the amount of recomposition

https://developer.android.com/jetpack/compose/performance#defer-reads

Programistich (approved):

PR #293, opened 2022-07-28
Fix incorrect rescale for nfc editor

LionZXY:
**Background**

Now we overlaps when screen is too small

**Changes**

- Use fontSize instead of width of each text field

**Test plan**

Try open screen on small phone

### DIFF
--- a/components/nfceditor/impl/src/main/java/com/flipperdevices/nfceditor/impl/composable/ComposableScaleFactorCalculate.kt
+++ b/components/nfceditor/impl/src/main/java/com/flipperdevices/nfceditor/impl/composable/ComposableScaleFactorCalculate.kt
@@ -16,6 +16,8 @@ import kotlin.math.roundToInt
 const val WIDTH_LINE_INDEX_DP = 9
 
 private const val NFC_LINE_BYTE_COUNT = 16
+
+// The accuracy with which we select the scaleFactor
 private const val SCALE_FACTOR_MULTIPLIER = 0.98f
 
 /**
--- a/components/nfceditor/impl/src/main/java/com/flipperdevices/nfceditor/impl/composable/ComposableScaleFactorCalculate.kt
+++ b/components/nfceditor/impl/src/main/java/com/flipperdevices/nfceditor/impl/composable/ComposableScaleFactorCalculate.kt
@@ -16,7 +16,7 @@ import kotlin.math.roundToInt
 const val WIDTH_LINE_INDEX_DP = 9
 
 private const val NFC_LINE_BYTE_COUNT = 16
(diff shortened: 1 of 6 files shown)


components/nfceditor/impl/src/main/java/com/flipperdevices/nfceditor/impl/composable/ComposableScaleFactorCalculate.kt
```diff
@@ -16,7 +16,7 @@ import kotlin.math.roundToInt
 const val WIDTH_LINE_INDEX_DP = 9
 
 private const val NFC_LINE_BYTE_COUNT = 16
-private const val SCALE_FACTOR_MULTIPLIER = 0.9f
+private const val SCALE_FACTOR_MULTIPLIER = 0.98f
```
Programistich, line 19: Maybe explain this magic number?

components/nfceditor/impl/src/main/java/com/flipperdevices/nfceditor/impl/composable/card/ComposableSchemeCard.kt
```diff
@@ -127,8 +133,8 @@ private fun ComposableWhiteSector() {
 
 @Composable
 @Suppress("LongMethod")
-private fun ComposableSecondSector() {
-    ComposableSector {
+private fun ComposableSecondSector(scaleFactor: Float) {
```
Programistich, line 136: NIT: what about put this parameter in CompositionLocal? Because we always provide that in all method

components/nfceditor/impl/src/main/java/com/flipperdevices/nfceditor/impl/composable/ComposableScaleFactorCalculate.kt
LionZXY: The accuracy with which we select the scaleFactor

components/nfceditor/impl/src/main/java/com/flipperdevices/nfceditor/impl/composable/card/ComposableSchemeCard.kt
LionZXY: Totally agree, but not now

components/nfceditor/impl/src/main/java/com/flipperdevices/nfceditor/impl/composable/ComposableScaleFactorCalculate.kt
Programistich: Yeah, it sounds good if we write a comment in code

PR #295, opened 2022-07-29
Deeplik open if activity exist

Programistich:
**Background**

We can't provide new deep link if application open

**Changes**

Process deeplink if activity exist
Log if we can't pop deeplink

**Test plan**

* Open app
* Going to some deep link
* In app you can see result on deeplink


LionZXY:
Good work, LGTM

PR #299, opened 2022-07-29
Migrate file manager to compose navigation

LionZXY:
**Background**

Now we use fragments and cicerone in file manager

**Changes**

- Use jetpack compose navigation

**Test plan**

Try navigate inside filemanager


components/settings/impl/src/main/java/com/flipperdevices/settings/impl/viewmodels/ExperimentalViewModel.kt
```diff
@@ -1,45 +1,29 @@
...
-        router.navigateTo(fileManager.fileManager())
+        navController.navigate(fileManagerEntry.fileManagerDestination())
     }
 
     fun onOpenScreenStreaming(navController: NavController) {
         metricApi.reportSimpleEvent(SimpleEvent.EXPERIMENTAL_OPEN_SCREENSTREAMING)
-        navController.navigate(NavGraphRoute.ScreenStreaming.name) {
-            popUpTo(NavGraphRoute.Settings.name)
-        }
+        navController.navigate(NavGraphRoute.ScreenStreaming.name)
```
Programistich, line 27: You dont remember pop to here?

Programistich (changes requested):

components/settings/impl/src/main/java/com/flipperdevices/settings/impl/viewmodels/ExperimentalViewModel.kt
LionZXY: What's it for? Everything works without it.

Programistich (approved):

PR #301, opened 2022-07-31
Add ability to download file from file manager

LionZXY:
**Background**

Now we can only upload to flipper

**Changes**

- Add ability to download from flipper file to android

**Test plan**

Try download file from flipper


Programistich:
Good work

Programistich (approved):

PR #302, opened 2022-07-31
Hex keyboard

Programistich:
**Background**

Hex keyboard for edit nfc dump

**Changes**

* Init new module
* Enum class with value
* Compose functions keyboard/key

**Test plan**

Run compose preview in module


components/core/ui/hexkeyboard/src/main/kotlin/com/flipperdevices/core/ui/hexkeyboard/ComposableHexKeyboard.kt
```diff
@@ -0,0 +1,109 @@
...
+import androidx.compose.runtime.CompositionLocalProvider
+import androidx.compose.runtime.compositionLocalOf
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.text.TextStyle
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+
+private val LocalButtonColor = compositionLocalOf<Color> { error("No button background") }
+private val LocalKeyAction = compositionLocalOf<(HexKey) -> Unit> { error("No key action") }
```
LionZXY, line 30: Please, add javadoc why we use localkeyaction

components/core/ui/hexkeyboard/src/main/kotlin/com/flipperdevices/core/ui/hexkeyboard/ComposableHexKeyboard.kt
```diff
@@ -0,0 +1,109 @@
...
+    onClick: (HexKey) -> Unit = {}
+) {
+    CompositionLocalProvider(
+        LocalTextStyle provides textStyle,
+        LocalContentColor provides contentColor,
+        LocalButtonColor provides backgroundKey,
+        LocalKeyAction provides onClick
+    ) {
+        Row(modifier.background(backgroundKeyBoard)) {
+            Column(Modifier.weight(1f)) {
```
LionZXY, line 48: Looks like the constraint layout is more usable in this case. Have you checked how it works with large print? With Constraint layout we can flatten the hierarchy and make the layout easier.

LionZXY (commented):
Globally, it's better to rewrite to Constraint layout. I can help you do that.

components/core/ui/hexkeyboard/src/main/kotlin/com/flipperdevices/core/ui/hexkeyboard/ComposableHexKeyboard.kt
```diff
@@ -0,0 +1,260 @@
...
+) {
+    CompositionLocalProvider(
+        LocalTextStyle provides textStyle,
+        LocalContentColor provides contentColor,
+        LocalButtonColor provides backgroundKey,
+        LocalKeyAction provides onClick
+    ) {
+        ConstraintLayout(modifier = modifier) {
+            val keys = HexKey.values()
+            val refs = keys.map { createRef() }
```
LionZXY, line 54: Let's use map EnumMap in this case

PR #303, opened 2022-07-31
Add editor to file manager

LionZXY:
**Background**

For now we can only download and share file, not edit

**Changes**

- Add ability edit files

**Test plan**

Try open text files and select edit

### DIFF
--- a/components/core/ui/theme/src/main/kotlin/com/flipperdevices/core/ui/theme/models/FlipperPallet.kt
+++ b/components/core/ui/theme/src/main/kotlin/com/flipperdevices/core/ui/theme/models/FlipperPallet.kt
@@ -98,7 +98,10 @@ data class FlipperPallet(
 
     val nfcCardBackground: Color = Color(0xFF4A4A4A),
     val nfcCardBackgroundSecond: Color = Color(0xFF303030),
-    val onNfcCard: Color = Color(0xFFFFFFFF)
+    val onNfcCard: Color = Color(0xFFFFFFFF),
+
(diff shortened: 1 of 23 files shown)


components/core/ui/theme/src/main/kotlin/com/flipperdevices/core/ui/theme/models/FlipperPallet.kt
```diff
@@ -98,7 +98,10 @@ data class FlipperPallet(
 
     val nfcCardBackground: Color = Color(0xFF4A4A4A),
     val nfcCardBackgroundSecond: Color = Color(0xFF303030),
-    val onNfcCard: Color = Color(0xFFFFFFFF)
+    val onNfcCard: Color = Color(0xFFFFFFFF),
+
+    val warningBackground: Color = Color(0xFFF63F3F),
```
Programistich, line 103: Warning for what? We have some colors for warning - error on try update/forgot flipper

Programistich (approved):

components/core/ui/theme/src/main/kotlin/com/flipperdevices/core/ui/theme/models/FlipperPallet.kt
LionZXY: Let's have some general color that alerts you to something, shall we?

LionZXY: Let's discuss this in the next PR

PR #306, opened 2022-08-01
Add ability to create file from file manager

LionZXY:
**Background**

Now we can edit, download and upload file

**Changes**

- Add ability to create file

**Test plan**

Try create file

### DIFF
--- a/components/core/ui/theme/src/main/kotlin/com/flipperdevices/core/ui/theme/models/FlipperPallet.kt
+++ b/components/core/ui/theme/src/main/kotlin/com/flipperdevices/core/ui/theme/models/FlipperPallet.kt
@@ -100,7 +100,7 @@ data class FlipperPallet(
     val nfcCardBackgroundSecond: Color = Color(0xFF303030),
     val onNfcCard: Color = Color(0xFFFFFFFF),
 
-    val warningBackground: Color = Color(0xFFF63F3F),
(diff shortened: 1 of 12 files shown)


components/core/ui/theme/src/main/kotlin/com/flipperdevices/core/ui/theme/models/FlipperPallet.kt
```diff
@@ -100,7 +100,7 @@ data class FlipperPallet(
     val nfcCardBackgroundSecond: Color = Color(0xFF303030),
     val onNfcCard: Color = Color(0xFFFFFFFF),
 
-    val warningBackground: Color = Color(0xFFF63F3F),
+    val warningColor: Color = Color(0xFFF63F3F),
```
Programistich, line 103: I think, we can set field with name for what warning

Programistich (approved):

components/core/ui/theme/src/main/kotlin/com/flipperdevices/core/ui/theme/models/FlipperPallet.kt
LionZXY: I suggest using a more generic color. We don't have the goal of making each part of the application a unique color, we have the goal of making it possible to apply themes. The danger/attention color will always be the same in these cases.

PR #310, opened 2022-08-03
Changelog on updater screen 

Programistich:
**Background**

Now we can see what's new in firmware

**Changes**

* Get color in fragment
* Library markdown
* Compose function for changeling
* Cancel button onPress effect
* Some changes padding

**Test plan**

Try update


LionZXY (approved):
Please, fix CI and then i merged it

PR #311, opened 2022-08-03
Migrate to faster build configuration

rougsig:
**Background**
Make faster builds.

**Changes**
1. Increase -Xms from 2g to 3g
5.2% (8.8s) of  build’s time was wasted to garbage collection. 
2. Replace all to bin gradle distribution. 
Just smaller jar, no docs, no sources.
3. Disable unnecessary android buildFeatures.
Less tasks = faster build
4. Disable viewBinding by default.
ViewBinding needed only for few modules. No need to enable this feature for all modules.
5. Disable Jetifier.
All project modules do not use any legacy support libs. (see [checkJetifierReport.txt](https://github.com/flipperdevices/Flipper-Android-App/files/9364391/checkJetifierReport.txt))

Should be faster more than 15%

**Test plan**
Green pipelines.


rougsig:
I'll add disable BuildConfig by default in another PR

PR #318, opened 2022-08-06
Not change state updater card after retry

Programistich:
**Background**

Now, if we have error on updater screen, we see error state after click retry

**Changes**

Setup state in progress after retry, because get information from flipper/fw server - its heavy

**Test plan**

..and how we can test it


LionZXY:
🔥

PR #319, opened 2022-08-06
Rewrite lottie component to custom

Programistich:
**Background**

On bottom bar freeze animation

**Changes**

* Replace Lottie on two images

**Test plan**

Open application and try connect


components/bottombar/api/src/test/kotlin/com/flipperdevices/bottombar/TabStateTest.kt
```diff
@@ -12,15 +12,19 @@ class TabStateTest {
         notSelectedIcon = 2,
         text = "TabState Animated 1",
         selectedColor = Color(0xFF111111),
-        unselectedColor = Color(0xFF222222)
+        unselectedColor = Color(0xFF222222),
+        selectedBackground = 0,
```
LionZXY, line 16: What is it?

components/bottombar/api/src/main/java/com/flipperdevices/bottombar/model/TabState.kt
```diff
@@ -53,9 +52,12 @@ sealed class TabState private constructor(
         }
     }
 
+    @Suppress("LongParameterList")
     class Animated(
-        @RawRes val selectedIcon: Int,
-        @RawRes val notSelectedIcon: Int,
+        @DrawableRes val selectedIcon: Int,
+        @DrawableRes val selectedBackground: Int,
+        @DrawableRes val notSelectedIcon: Int,
```
LionZXY, line 59: Maybe mark it as `null`?

Programistich: In what case its help?

components/bottombar/api/src/test/kotlin/com/flipperdevices/bottombar/TabStateTest.kt
Programistich: It's test, I think necessary in test

PR #323, opened 2022-08-07
Sync key if change name key

Programistich:
**Background**

Now we edit description and key stay unsync

**Changes**

Check if name key equals

**Test plan**

Edit description and see key was sync


components/keyedit/impl/src/main/java/com/flipperdevices/keyedit/impl/viewmodel/KeyEditViewModel.kt
```diff
@@ -106,11 +106,16 @@ class KeyEditViewModel(
             ),
             flipperKey.keyContent,
             editState.notes,
-            synchronized = false
+            // revert, because now its not synchronize
+            synchronized = !isNeedSynchronization(editState)
```
LionZXY, line 110: You need change logic in database save, not in UI

LionZXY (changes requested):

PR #326, opened 2022-08-09
Update manifest for 3rd file manager by android 11+

Programistich:
**Background**

Now we can't use 3rd file manager for upload file on android 11+

**Changes**

Update manifest for GetContent

**Test plan**

Try upload file from 3rd file picker


LionZXY:
Good research, thank you!

PR #333, opened 2022-08-16
Fix wrong counter when reporting iButton sync end

coletz:
**Background**

Found a possible copy-paste issue while checking the code

**Changes**

Fixed wrong enum used

**Test plan**

I don't know, code just makes sense


LionZXY:
Thanks for your attention!

PR #343, opened 2022-08-29
Not exist region file v2

Programistich:
**Background**

Suggest update flipper if region file not exist

**Changes**

* FileExistHelper 
* Tests

**Test plan**

* Delete region file and recheck updates


LionZXY:
It looks fantastic!

PR #345, opened 2022-08-29
Offer update when region changes after last update

Programistich:
**Background**

 Offer update when region changes after last update

**Changes**

* Remove double divider
* Bumb markdown version
* Method for SubGhz provisioning for get region (Maybe any idea for great solution) + test
* UpdateCardVM check if region change + test
* Save region in setting proto

**Test plan**

Travel in another region, because I cant :D


components/core/preference/src/main/proto/settings.proto
```diff
@@ -37,4 +37,5 @@ message Settings {
   SelectedTheme selected_theme = 13;
   bool expert_mode = 14;
   bool ignore_subghz_provisioning_on_zero_region = 15;
+  string region = 16;
```
LionZXY, line 40: Wouldn't it be better to rename it to something like `last_provided_region`?

components/updater/api/src/main/java/com/flipperdevices/updater/api/SubGhzProvisioningHelperApi.kt
```diff
@@ -0,0 +1,8 @@
+package com.flipperdevices.updater.api
+
+import com.flipperdevices.bridge.service.api.FlipperServiceApi
+
+interface SubGhzProvisioningHelperApi {
```
LionZXY, line 5: It looks like it should be moved to a separate module now - the current module is too overloaded.

components/updater/card/src/main/java/com/flipperdevices/updater/card/viewmodel/UpdateCardViewModel.kt
```diff
@@ -140,8 +142,9 @@ class UpdateCardViewModel @VMInject constructor(
             isManifestExist(serviceApi),
             isSubGhzProvisioningExist(serviceApi)
         ) { setting, isManifestExist, isSubGhzProvisioningExist ->
-            // TODO check previous region
-            return@combine setting.alwaysUpdate || !isManifestExist || !isSubGhzProvisioningExist
+            val isRegionChanged = subGhzProvisioningHelperApi.getRegion() != setting.region
+            return@combine setting.alwaysUpdate || !isManifestExist ||
+                !isSubGhzProvisioningExist || isRegionChanged
```
LionZXY, line 147: It looks as if this class should be split into several.

components/updater/impl/src/main/java/com/flipperdevices/updater/impl/tasks/SubGhzProvisioningHelper.kt
```diff
@@ -38,19 +39,23 @@ import kotlinx.coroutines.withTimeoutOrNull
...
     private val downloaderApi: DownloaderApi,
     private val regionProvisioningHelper: RegionProvisioningHelper,
     private val metricApi: MetricApi,
     private val settings: DataStore<Settings>
-) : SubGhzProvisioningHelper, LogTagProvider {
+) : SubGhzProvisioningHelperApi, LogTagProvider {
     override val TAG = "SubGhzProvisioningHelper"
 
+    override suspend fun getRegion(): String? {
+        val response = downloaderApi.getSubGhzProvisioning()
```
LionZXY, line 52: What happens if there is no network? How will this method and the callers react?

gradle/libs.versions.toml
```diff
@@ -40,7 +40,7 @@ appcompat = "1.5.0"
 datastore = "1.0.0"
 # https://github.com/vsch/flexmark-java/issues/442
 flexmark = "0.42.14"
-markdown = "0.1.2"
+markdown = "0.1.5"
```
LionZXY, line 43: Please don't mix different features into one. Bring it back.

components/updater/impl/src/main/java/com/flipperdevices/updater/impl/tasks/SubGhzProvisioningHelper.kt
```diff
@@ -68,9 +73,11 @@ class SubGhzProvisioningHelperImpl @Inject constructor(
             response.countries[providedRegion.uppercase()] ?: response.defaults
         } else response.defaults
 
+        val finalCodeRegion = providedRegion?.uppercase() ?: UNKNOWN_REGION
```
LionZXY, line 76: Why don't you use getRegion here?

LionZXY (approved):
Good work

PR #348, opened 2022-08-31
Display error on updater screen if /int/ full

Programistich:
**Background**

If Flipper has full internal storage, it rebooting

**Changes**

* Compare with error code
* Exception when internal storage full
* State `Internal full`
* addition text style markdown for apply to all styles(for example center all)

**Test plan**

Flash doom on our flipper and try update from phone


components/updater/impl/src/main/java/com/flipperdevices/updater/impl/UpdaterTask.kt
```diff
@@ -134,8 +135,10 @@ class UpdaterTask(
             )
         } catch (e: Throwable) {
             error(e) { "Failed when upload to flipper" }
-            if (e !is CancellationException) {
-                stateListener(UpdatingState.FailedUpload)
+            when (e) {
+                is IntFlashFullException -> stateListener(UpdatingState.FailedInternalStorage)
+                is CancellationException -> {}
+                else -> stateListener(UpdatingState.FailedUpload)
```
LionZXY, line 141: Good code

components/updater/api/src/main/java/com/flipperdevices/updater/model/UpdatingState.kt
```diff
@@ -13,6 +13,7 @@ sealed class UpdatingState(
 ) {
     object NotStarted : UpdatingState(true)
 
+    object FailedInternalStorage : UpdatingState(true)
```
LionZXY, line 16: It's a good decision to create a separate state

LionZXY (approved):

PR #350, opened 2022-09-01
Restart rpc on error

LionZXY:
**Background**

Right now we just crash if we see an error with RPC.

**Changes**

- Restart RPC if an error occurs with protobuf parsing
- Restart the RPC if the flipper does not respond for a long time

**Test plan**

Try press to Restart RPC button in debug settings


Programistich (approved):
Good Bye .dev packages :(((

PR #360, opened 2022-09-02
Fix Gradle configuration and implement checksums for security

BoBeR182:
**Background**

https://gitlab.com/fdroid/rfp/-/issues/2187#note_1084842373

gradle/wrapper/gradle-wrapper.jar is [gradle-7.3-rc-3-wrapper.jar](https://services.gradle.org/distributions/gradle-7.3-rc-3-wrapper.jar.sha256), but [gradle/wrapper/gradle-wrapper.properties](https://github.com/flipperdevices/Flipper-Android-App/blob/f972d9784fda357d627d9aa1d42d835766871189/gradle/wrapper/gradle-wrapper.properties) declares https://services.gradle.org/distributions/gradle-7.5.1-bin.zip as the gradle version.

gradle/wrapper/gradle-wrapper.properties is missing [distributionSha256Sum](https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:verification).

This means that the gradle download is not verified.
We recommend explicitly setting the expected Sha256sum to protect you and your apps if a bad actor gets access to the Gradle servers or manages to MitM your internet connection.
Please note that Android Studio does not currently support this and may or may  not crash, but we suggest you try our suggestion anyway to improve the security of yourself and your app.

**Changes**

fixed above issues by updating binary and adding checksum

**Test plan**

pass pipeline, compile apk,


LionZXY:
Thank you!

PR #371, opened 2022-09-15
Fix synchronization status with shadow files

LionZXY:
**Background**

Right now we have bugs with synchronization and shadow files

**Changes**

- Create a shadow file when editing a card even if there is no shadow file
- Mark all keys that are present at the moment as synchronized

**Test plan**
Try renaming the file back and forth. For example, "Office.nfc"->"Office_1.nfc" -> "Office.nfc".
Behavior with old firmware: never sync
Behavior with new firmware: synchronizes


Programistich (approved):
Good work!

PR #373, opened 2022-09-15
Warning when storage ending

Programistich:
**Background**

Change text color in info stat when storage(int/ext) ending

**Changes**

* Geminio config update for test
* Changes color text in info
* DeviceInfo method for check less 20Kb + TESTS🌚
* Fix get color from branch, not commit(fun getFirmwareChannel(commit: String?): FirmwareChannel?)

**Test plan**

Try install doom


components/info/impl/src/main/java/com/flipperdevices/info/impl/model/DeviceInfo.kt
```diff
@@ -7,15 +7,26 @@ import com.flipperdevices.info.impl.R
 import com.flipperdevices.updater.model.FirmwareVersion
 import kotlin.math.max
 
+private const val ENDING = 20 * 1000 // 20 Kb
```
LionZXY, line 10: Maybe use 1024?

PR #374, opened 2022-09-16
Import full device info for support

Programistich:
**Background**

Now we can share full device info to another application

**Changes**

* New module `share` for sharing file
* FullInfoViewModel - logic for full info screen
* PowerInfo in rpc information

**Test plan**

Go to Full Info and click upload in right top side


components/core/share/src/main/kotlin/com/flipperdevices/core/share/ShareHelper.kt
```diff
@@ -0,0 +1,28 @@
+package com.flipperdevices.core.share
+
+import android.content.Context
+import android.content.Intent
+import androidx.core.content.FileProvider
+import java.io.File
+
+object ShareHelper {
+    fun shareFile(context: Context, file: File, name: String, resId: Int) {
```
LionZXY, line 9: You should check folder of file. And if folder is not sharedkeys, copy it or throw exception

components/info/impl/src/main/java/com/flipperdevices/info/impl/viewmodel/FullInfoViewModel.kt
```diff
@@ -0,0 +1,135 @@
...
+) : AndroidLifecycleViewModel(application),
+    FlipperBleServiceConsumer,
+    LogTagProvider {
+    override val TAG = "FullInfoViewModel"
+    private val deviceInfoState = MutableStateFlow(DeviceInfo())
+    private val flipperRpcInformationState = MutableStateFlow(FlipperRpcInformation())
+    private val deviceInfoRequestStatus = MutableStateFlow(DeviceInfoRequestStatus())
+
+    init {
+        ComponentHolder.component<InfoComponent>().inject(this)
```
LionZXY, line 51: Why you use inject with vminject? You should delete them

components/info/impl/src/main/java/com/flipperdevices/info/impl/viewmodel/FullInfoViewModel.kt
```diff
@@ -0,0 +1,135 @@
...
+        try {
+            addInfoToFile(file, context)
+            ShareHelper.shareFile(
+                context = context,
+                file = file,
+                name = nameFile,
+                resId = R.string.device_info_share
+            )
+        } catch (@Suppress("SwallowedException") exception: Exception) {
+            error { "Exception when upload device info" }
```
LionZXY, line 110: You shouldn't suppress exception, print it

components/info/impl/src/main/java/com/flipperdevices/info/impl/viewmodel/FullInfoViewModel.kt
```diff
@@ -0,0 +1,135 @@
...
+        val flipperName = flipperRpcInformationState.value.flipperDeviceInfo.deviceName ?: "unknown"
+        val formatDate = LocalDateTime.now().format(
+            DateTimeFormatter.ofPattern("yyyy-MM-dd-HH-mm-ss")
+        )
+        return "dump-$flipperName-$formatDate.txt"
+    }
+
+    private fun addInfoToFile(file: File, context: Context) {
+        flipperRpcInformationState.value.allFields.forEach { (t, u) ->
+            file.appendNewLine("$t:$u")
```
LionZXY, line 124: NIT: Maybe use strinbuilder here and then print all this string in file?

LionZXY (changes requested):

LionZXY (approved):

LionZXY (approved):

LionZXY (approved):

PR #388, opened 2022-09-23
Fix name of updater folder

Programistich:
**Background**

Duplicate folder for release and rc builds
Short name get from folder, not url

**Changes**

Get internal folder name

**Test plan**

Upload to flipper rc build
Upload to flipper release build
Check what diff folder on flipper

### DIFF
--- a/components/updater/impl/src/main/java/com/flipperdevices/updater/impl/UpdaterTask.kt
+++ b/components/updater/impl/src/main/java/com/flipperdevices/updater/impl/UpdaterTask.kt
@@ -151,12 +151,12 @@ class UpdaterTask(
     ): String {
         flipperUpdateImageHelper.loadImageOnFlipper(requestApi)
 
-        val internalFolder = updaterFolder
+        val updateName: String = updaterFolder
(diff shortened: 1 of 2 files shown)


components/updater/impl/src/main/java/com/flipperdevices/updater/impl/UpdaterTask.kt
```diff
@@ -147,14 +146,17 @@ class UpdaterTask(
...
-        var updateName = updateFile.url.substringAfterLast("/").substringBefore(".")
-        if (updateName.isBlank()) {
-            updateName = updateFile.sha256
-        }
         flipperUpdateImageHelper.loadImageOnFlipper(requestApi)
+
+        val internalFolder = updaterFolder
+            .listFiles()
+            ?.first { it.isDirectory }
+            ?: error("Not found folder in updaterFolder on prepare upload")
```
LionZXY, line 157: You don't have to throw an error here, let's just use the default name?

LionZXY (approved):

PR #390, opened 2022-09-24
Redesign card when we waiting update

Programistich:
**Background**

Redesign card when we waiting update 

**Changes**

Check updating status

**Test plan**

Update Flipper and check phone after reboot device


components/info/impl/build.gradle.kts
```diff
@@ -19,6 +19,7 @@ dependencies {
     implementation(projects.components.core.ui.lifecycle)
     implementation(projects.components.core.ui.theme)
     implementation(projects.components.core.share)
+    implementation(projects.components.updater.card)
```
LionZXY, line 22: Use only `api` module always

LionZXY (approved):

LionZXY:
Thank you!

PR #391, opened 2022-09-24
Unknown percent on error update

Programistich:
**Background**

Now, if firmware not exist on server(or some trouble with server) we have total bytes -1

**Changes**

Check what total bytes not less 0

**Test plan**

### DIFF
--- a/components/updater/downloader/src/main/java/com/flipperdevices/updater/downloader/api/DownloaderApiImpl.kt
+++ b/components/updater/downloader/src/main/java/com/flipperdevices/updater/downloader/api/DownloaderApiImpl.kt
@@ -22,7 +22,6 @@ import io.ktor.client.HttpClient
 import io.ktor.client.call.body
 import io.ktor.client.request.get
 import java.io.File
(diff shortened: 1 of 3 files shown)


LionZXY (approved):

components/updater/downloader/src/main/java/com/flipperdevices/updater/downloader/api/DownloaderApiImpl.kt
```diff
@@ -119,7 +119,9 @@ class DownloaderApiImpl @Inject constructor(
                     distributionFile,
                     tempFile
                 ) { processedBytes, totalBytes ->
-                    send(DownloadProgress.InProgress(processedBytes, totalBytes))
+                    if (totalBytes > 0) {
+                        send(DownloadProgress.InProgress(processedBytes, totalBytes))
+                    }
```
LionZXY, line 124: What are we showing the user here? Shouldn't we throw out the error?

LionZXY (approved):

PR #398, opened 2022-10-02
Get last region from flipper

Programistich:
**Background**

Get last region from flipper and compare with current on phone

**Changes**

Get file .region_data and take country code

**Test plan**

Green pipeline

### DIFF
--- a/components/bridge/api/src/main/java/com/flipperdevices/bridge/api/model/FlipperRequest.kt
+++ b/components/bridge/api/src/main/java/com/flipperdevices/bridge/api/model/FlipperRequest.kt
@@ -1,10 +1,11 @@
 package com.flipperdevices.bridge.api.model
 
+import com.flipperdevices.core.ktx.jre.TimeHelper
 import com.flipperdevices.protobuf.Flipper
 
 data class FlipperRequest(
     val data: Flipper.Main,
-    val createTimestampNanos: Long = System.nanoTime(),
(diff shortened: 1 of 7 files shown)


components/bridge/api/src/main/java/com/flipperdevices/bridge/api/model/FlipperRequest.kt
```diff
@@ -7,7 +7,22 @@ data class FlipperRequest(
...
+        other as FlipperRequest
+
+        if (data != other.data) return false
+
+        return true
+    }
+
+    override fun hashCode(): Int {
+        return data.hashCode()
+    }
```
LionZXY, line 24: Why did you add that? Perhaps you forgot to remove it?

components/core/preference/src/main/proto/settings.proto
```diff
@@ -37,5 +37,4 @@ message Settings {
   SelectedTheme selected_theme = 13;
   bool expert_mode = 14;
   bool ignore_subghz_provisioning_on_zero_region = 15;
-  string last_provided_region = 16;
```
LionZXY, line 40: Please, never delete protobuf field without reserved. In future when we add new field with this id crash happens

LionZXY (changes requested):

LionZXY (approved):

PR #399, opened 2022-10-03
Flipper widget part 1

LionZXY:
**Background**

Widgets require logic and design - I decided to break this down into two or more PRs.

**Changes**

- Add widget to new module
- Add designs for widgets

**Test plan**

Try add widget to your homescreen


components/widget/impl/src/main/java/com/flipperdevices/widget/impl/di/ImplComponent.kt
```diff
@@ -0,0 +1,7 @@
+package com.flipperdevices.widget.impl.di
+
+import com.flipperdevices.core.di.AppGraph
+import com.squareup.anvil.annotations.ContributesTo
+
+@ContributesTo(AppGraph::class)
+interface ImplComponent
```
Programistich, line 7: What about WidgetComponent?

Programistich (changes requested):
Good work

PR #402, opened 2022-10-05
Add key picker for widget

LionZXY:
**Background**

Right now we can create widget, but can't pick key for widget

**Changes**

Add widget picker screen

**Test plan**

- Create widget
- Open widget settings
- Try search


components/singleactivity/impl/src/main/AndroidManifest.xml
```diff
@@ -4,9 +4,9 @@
     <application>
         <activity
                 android:exported="false"
-                android:screenOrientation="portrait"
                 android:launchMode="singleTask"
                 android:name=".SingleActivity"
+                android:screenOrientation="portrait"
```
Programistich, line 9: NIT:Maybe revert that?

Programistich (changes requested):
Placeholder in Choose file
Nothing do after choose key

Programistich (approved):

PR #403, opened 2022-10-05
Update from web updater

Programistich:
**Background**

Update from web updater

**Changes**

* Deeplink for parse query (Not use FFFDecoder because that user fragment from uri)
* Deeplink handler navigate to updater screen 

**Test plan**

### DIFF
--- a/components/deeplink/impl/src/main/java/com/flipperdevices/deeplink/impl/parser/delegates/DeepLinkWebUpdater.kt
+++ b/components/deeplink/impl/src/main/java/com/flipperdevices/deeplink/impl/parser/delegates/DeepLinkWebUpdater.kt
@@ -2,7 +2,6 @@ package com.flipperdevices.deeplink.impl.parser.delegates
 
 import android.content.Context
 import android.content.Intent
-import android.net.Uri
 import com.flipperdevices.core.di.AppGraph
 import com.flipperdevices.core.log.LogTagProvider
 import com.flipperdevices.deeplink.api.DeepLinkParserDelegate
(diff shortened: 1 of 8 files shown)


components/deeplink/impl/src/main/java/com/flipperdevices/deeplink/impl/parser/delegates/DeepLinkWebUpdater.kt
```diff
@@ -0,0 +1,49 @@
...
+        if (intent.data?.host == HOST) {
+            return DeepLinkParserDelegatePriority.HIGH
+        }
+        return DeepLinkParserDelegatePriority.LAST_CHANCE
+    }
+
+    override suspend fun fromIntent(context: Context, intent: Intent): Deeplink? {
+        val uri = intent.data ?: return null
+        val content = uriToContent(uri)
+        val link = content.find { it.first == "url" }?.second ?: return null
```
LionZXY, line 36: Please move it to const

components/deeplink/impl/src/main/java/com/flipperdevices/deeplink/impl/parser/delegates/DeepLinkWebUpdater.kt
```diff
@@ -0,0 +1,49 @@
...
+        val version = content.find { it.first == "version" }?.second ?: ""
+        val channel = content.find { it.first == "channel" }?.second ?: ""
+        return Deeplink.WebUpdate(link, "$channel $version")
+    }
+
+    private fun uriToContent(uri: Uri): List<Pair<String, String>> {
```
LionZXY, line 42: Please use inbuild solution in `uri`

PR #406, opened 2022-10-07
Add delete action in filemanager

LionZXY:
**Background**

Right now we cannot delete files from the file manager

**Changes**

- Added item to the dialog - delete

**Test plan**

Try delete file by file manager


Programistich (commented):
What about delete /int/? its unreal?

PR #410, opened 2022-10-12
Delete connection request when disconnect called

LionZXY:
**Background**

Right now when we press disconnect button when connection happening queue is full and we don't call disconnect

**Changes**

- Clear request queue and cancel connection when call disconnect

**Test plan**

Press connect with disabled flipper
Press disconnect then


Programistich (dismissed):
Good work

PR #411, opened 2022-10-12
Add selection in device info cards

LionZXY:
**Background**

Right now we can't select text in device info screen

**Changes**

- Wrap card in selection container

**Test plan**

Try select text in device info screen


Programistich (changes requested):
How about change color cursor

PR #425, opened 2022-10-26
Updater card display update from url

Programistich:
**Background**

Set update in card

**Changes**

* Set deeplink in all fragment
* Handle intent for Compose Navigation
* 

**Test plan**

* Try click on FW repository to Web/App updater
* Opening application and in Updater Card see web update


LionZXY (approved):
Excelent work!

PR #439, opened 2022-11-14
New share flow

Programistich:
**Background**

Now we share sooo long key link
On new share flow we can detect long link and share shortlink (optional file)

**Changes**

*  Bottom sheet uploader
    * Calculate default link, if more 200 symbols - share by server
    * Share by file - get stream from key
    * Encode/Decode key by AES
    * Send/Download file by array,name/id to our server
    * Error states + button retry
* Crypto deeplink
    *  Parse deeplink - find key/path/fileID
    *  Parse deeplink crypto link in RecieveFragment
    *  Error states + button retry

**Test plan**
* Open big key - click share - click shre by link - wait - send to messages
* Click on link in message -> opened Add Key fragment -> add to archive


components/bridge/dao/api/src/main/java/com/flipperdevices/bridge/dao/api/model/FlipperKey.kt
```diff
@@ -22,11 +24,15 @@ data class FlipperKey(
         get() = mainFile.content
 
     fun getKeyPath() = FlipperKeyPath(mainFile.path, deleted)
+    fun getKeyContent() = mainFile.content.openStream().use { it.readBytes() }
```
LionZXY, line 27: Please, don't add it in general class

LionZXY (approved):

PR #464, opened 2022-12-21
New metric for share flow

Programistich:
**Background**
New metric for share flow

**Changes**
* New simple event value
* Call metric in share flow

**Test plan**
Green pipeline


LionZXY (approved):
Good work

PR #465, opened 2022-12-21
New errors type on import key

Programistich:
**Background**

When we import a key, we may get errors

**Changes**

* Different errors on share and receive
* Exception 404 when key not found
* Icons for errors

**Test plan**
Green pipeline


components/share/receive/src/main/java/com/flipperdevices/share/receive/composable/screen/ComposableKeyErrorScreen.kt
```diff
@@ -61,52 +61,67 @@ private fun ComposableErrorContent(typeError: ShareContentError, onRetry: () ->
...
+
+        val isDisplayRetry = isCanUserRetryImportKey(typeError)
+        if (isDisplayRetry) {
+            Text(
+                modifier = Modifier
+                    .padding(top = 6.dp)
+                    .clickable(
+                        indication = rememberRipple(),
+                        onClick = onRetry,
+                        interactionSource = remember { MutableInteractionSource() }
```
LionZXY, line 73: Good idea use remember here. Let's move this method in ktx module and use it in whole project

LionZXY (approved):

PR #471, opened 2022-12-22
Fix crash when delete from trash key with favorites

LionZXY:
**Background**

Right now when we delete key from trash and this key exist in favorites, app throw crash

**Changes**

Delete key before permanently delete from trash

**Test plan**

- Add key to favorite
- Move it to trash
- Delete it from trash


Programistich (approved):
Nice! Good job

PR #473, opened 2022-12-23
Refactor `share:receive` module

Programistich:
**Background**

Clear code

**Changes**
* Placeholder: custom modifier with wight/height element; flag for display placeholder
*  Step to move fragments and use dependencies from viewmodel
* Internal Composable function
* Refactor naming folder
* ReceiveKeyActionHelper - use case for action on key

**Test plan**

Green pipeline and qa testing


components/share/receive/src/main/java/com/flipperdevices/share/receive/composable/ComposableKeyReceive.kt
```diff
@@ -1,42 +1,43 @@
...
+            typeError = localState.type,
             onCancel = onCancel,
             onRetry = viewModel::onRetry
         )
-        ReceiveState.Finished -> return
+        ReceiveState.NotStarted -> ComposableKeyInProgressScreen(
+            keyScreenApi = keyScreenApi,
+            onCancel = onCancel
+        )
+        ReceiveState.Finished -> onFinish()
```
LionZXY, line 41: Use lambda here please

LionZXY (approved):
Good work

PR #503, opened 2023-01-19
Key encryption without paddings

Programistich:
**Background**

Key encryption without paddings

**Changes**

* Use encoder without padding
* Replace id/path in link

**Test plan**

Try generate share link on android and open on ios


components/share/cryptostorage/src/main/kotlin/com/flipperdevices/share/cryptostorage/helper/CryptoHelper.kt
```diff
@@ -30,7 +30,11 @@ class CryptoHelperApiImpl @Inject constructor() : CryptoHelperApi {
             .getInstance(KeyProperties.KEY_ALGORITHM_AES)
             .apply { init(KEY_SIZE) }
         val secretKey = generator.generateKey()
-        val keyString = Base64.getUrlEncoder().encodeToString(secretKey.encoded)
+
+        val keyString = Base64
+            .getUrlEncoder()
+            .withoutPadding()
```
LionZXY, line 36: Awesome! Great work!

LionZXY (approved):

PR #504, opened 2023-01-20
Refactor synchronization

LionZXY:
**Background**

Right now we can only synchronize all keys

**Changes**

Add per-folder synchronization

**Test plan**

Try synchronize folders


Programistich (approved):
Good Work

PR #506, opened 2023-01-23
On `save dump as` save only editable key

Programistich:
**Background**

On `save dump as` save only editable key

**Changes**

* Method for new content by state + key
* Method add key with change content in shadow file
* Method return new key by new content

**Test plan**

* Change some data and `save dump as` 
* Go to key file and see changed data


LionZXY (approved):

LionZXY (approved):
Thank you for tests! ❤️

LionZXY (approved):

PR #508, opened 2023-01-25
[Draft] Cipher stream for crypto operation

Programistich:
**Background**

Cipher stream for crypto operation

**Changes**

* Use cipher stream

**Test plan**

Try share and import key

### DIFF
--- a/components/bridge/dao/api/src/main/java/com/flipperdevices/bridge/dao/api/model/FlipperKey.kt
+++ b/components/bridge/dao/api/src/main/java/com/flipperdevices/bridge/dao/api/model/FlipperKey.kt
@@ -24,6 +24,10 @@ data class FlipperKey(
         get() = mainFile.content
 
     fun getKeyPath() = FlipperKeyPath(mainFile.path, deleted)
+
+    fun getShadowFile(): FlipperFile? {
+        return additionalFiles.firstOrNull { it.path.fileType == FlipperFileType.SHADOW_NFC }
+    }
 }
 
 @Parcelize
--- a/components/share/cryptostorage/src/main/kotlin/com/flipperdevices/share/cryptostorage/CryptoStorageApiImpl.kt
+++ b/components/share/cryptostorage/src/main/kotlin/com/flipperdevices/share/cryptostorage/CryptoStorageApiImpl.kt
@@ -1,6 +1,8 @@
 package com.flipperdevices.share.cryptostorage
 
 import com.flipperdevices.bridge.dao.api.delegates.KeyParser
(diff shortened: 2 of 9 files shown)


components/bridge/dao/api/src/main/java/com/flipperdevices/bridge/dao/api/model/FlipperKey.kt
```diff
@@ -24,6 +24,10 @@ data class FlipperKey(
         get() = mainFile.content
 
     fun getKeyPath() = FlipperKeyPath(mainFile.path, deleted)
+
+    fun getShadowFile(): FlipperFile? {
```
LionZXY, line 28: As I see it, this is only used in one place. I suggest removing it, since we can't add the same method for each file type.
For example, if we have additional files, we won't create another method in a common class, will we?

components/share/cryptostorage/src/main/kotlin/com/flipperdevices/share/cryptostorage/CryptoStorageApiImpl.kt
```diff
@@ -15,23 +17,23 @@ class CryptoStorageApiImpl @Inject constructor(
     private val storageHelperApi: StorageHelperApi,
     private val keyParser: KeyParser
 ) : CryptoStorageApi {
-    override suspend fun upload(data: ByteArray, path: String, name: String): Result<String> {
+    override suspend fun upload(flipperKey: FlipperKey): Result<String> {
         return runCatching {
-            val encryptedData = cryptoHelperApi.encrypt(data)
+            val flipperFile = flipperKey.getShadowFile() ?: flipperKey.mainFile
```
LionZXY, line 22: Also, please, add check for NFC type

components/share/cryptostorage/src/main/kotlin/com/flipperdevices/share/cryptostorage/helper/CryptoHelper.kt
```diff
@@ -43,12 +49,23 @@ class CryptoHelperApiImpl @Inject constructor() : CryptoHelperApi {
...
+
+        inputStream.use { input ->
+            cipherStream.use { output ->
+                input.copyTo(output)
+            }
+        }
+
         val iv = encryptionCipher.iv
-        val encryptedBytes = encryptionCipher.doFinal(data)
+        val encryptedBytes = outputStream.toByteArray()
```
LionZXY, line 64: You should rewrite `CryptoHelperApiImpl` to stream api, because in this case you also store whole bytes in memmory

PR #513, opened 2023-01-31
Migrate bottombar to Jetpack Compose

LionZXY:
**Background**

Right now we migrate separate parts of application to Jetpack Compose, but not whole

**Changes**

- Migrate BottomBar module to jetpack compose
- Migrate SingleActivity to Jetpack Compose


components/bottombar/impl/src/main/java/com/flipperdevices/bottombar/impl/viewmodel/BottomNavigationViewModel.kt
```diff
@@ -0,0 +1,76 @@
...
+    override val TAG = "BottomNavigationViewModel"
+
+    private val selectedTabInternal = MutableStateFlow(FlipperBottomTab.DEVICE)
+
+    val selectedTab: StateFlow<FlipperBottomTab>
+        get() = selectedTabInternal
+
+    fun getStartDestination(): FlipperBottomTab {
+        return runBlockingWithLog("selected_tab") {
+            val selectedTab = settingsDataStore.data.first().selectedTab
```
Programistich, line 32: WIP. We can write val selectedTab = settingsDataStore.data.first().selectedTab and after that check on null in when

Programistich (changes requested):
GOOD JOB!

Programistich (approved):

Programistich (approved):

PR #523, opened 2023-02-10
Remove fragements from project

Programistich:
**Background**

Remove fragements from project

**Changes**

* Remove cicerone library
* Remove core/navigation module

**Test plan**

Try clicks all flows on app and green pipeline


LionZXY (approved):
🔥

PR #526, opened 2023-02-17
Migrate to new device info api

LionZXY:
**Background**

Right now we use deprecated API for system info 

**Changes**

- Move rpc information from bridge module
- Not execute rpc information on each connection
- Support old and new api for rpc information
- Not execute rpc information if not open full info

**Test plan**

Try open full info and device info


components/analytics/shake2report/impl/build.gradle.kts
```diff
@@ -12,6 +12,7 @@ dependencies {
     implementation(projects.components.core.di)
     implementation(projects.components.core.activityholder)
     implementation(projects.components.core.preference)
+
```
Programistich, line 15: necessary line

components/bridge/rpcinfo/api/src/main/java/com/flipperdevices/bridge/rpcinfo/model/FlipperStorageInformation.kt
```diff
@@ -0,0 +1,57 @@
...
+import kotlin.math.max
+
+data class FlipperStorageInformation(
+    val internalStorageStatus: FlipperInformationStatus<StorageStats?> =
+        FlipperInformationStatus.NotStarted(),
+    val externalStorageStatus: FlipperInformationStatus<StorageStats?> =
+        FlipperInformationStatus.NotStarted()
+)
+
+fun <T> FlipperInformationStatus<T>.dataOrNull() = when (this) {
```
Programistich, line 15: How about set this method inside FlipperInformationStatus?

Programistich (changes requested):

components/bridge/rpcinfo/api/src/main/java/com/flipperdevices/bridge/rpcinfo/model/FlipperStorageInformation.kt
LionZXY: It was important for me to separate the data from the data processing methods

PR #527, opened 2023-03-01
Migrate VersionApi to isSupported

LionZXY:
**Background**

Right now we each time request version and wait until version is provided

**Changes**

- Share this logic in version api class
- Use isSupported method 

**Test plan**

Try launch application and check version support


Programistich (approved):
Good work

PR #531, opened 2023-03-01
New report bug screen

LionZXY:
**Background**

Right now we use deprecated shake2report interface with activity

**Changes**

- Migrate shake2report to compose
- New UI with name and description

**Test plan**

Try report test bug


Programistich (approved):
Thx for that

PR #533, opened 2023-03-02
Fix theme change 

LionZXY:
**Background**

Right now we have empty screen on theme changing

**Changes**

- Move all application-scope action in root application class (like metric report or start synchronization)
- Remove condition which clear activity when activity reconfigurate
- Add animation on color changing
- Move all default variable in FlipperPallet body

**Test plan**

Try change theme


Programistich (dismissed):

Programistich (approved):
Good work

PR #539, opened 2023-03-06
New report bug screen v2 

Programistich:
**Background**

New report bug screen v2

**Changes**

Screen by design

**Test plan**

Try report bug


components/analytics/shake2report/impl/src/main/java/com/flipperdevices/analytics/shake2report/impl/viewmodel/Shake2ReportViewModel.kt
```diff
@@ -0,0 +1,122 @@
...
+            return@launch
+        }
+
+        runCatching {
+            shake2ReportStateFlow.emit(Shake2ReportState.Uploading)
+            val compressingFile = if (addLogs) compressLogFolder() else null
+            val sentryId = sendingReport(name, description, compressingFile)
+            shake2ReportStateFlow.emit(Shake2ReportState.Complete(sentryId))
+        }.onFailure {
+            shake2ReportStateFlow.emit(Shake2ReportState.Error)
```
LionZXY, line 60: Could you, please, also add error on timeout flush?
I am don't know possible it or not, so you need do some investigation

LionZXY (approved):

PR #542, opened 2023-03-18
Bump deps

LionZXY:
**Background**

Right now we have deprecated deps

**Changes**

Bump deps

**Test plan**

Full regress


Programistich (approved):
Thank You!

PR #543, opened 2023-03-21
Self update by google play service

Programistich:
**Background**

Suggest update application by Google Play Core

**Changes**

* InAppNotification new data model - saved key and new update
* Composable function for up point
* selfupdater module like api - impl(google, github, none)
* SelfUpdaterGooglePlayImpl with realization suggestion 

**Test plan**

* Publish that version - X
* Build next version by that - Y
* Go to install version X and wait for suggest Y update


components/bottombar/impl/src/main/java/com/flipperdevices/bottombar/impl/api/BottomNavigationFeatureEntryImpl.kt
```diff
@@ -49,6 +53,18 @@ class BottomNavigationFeatureEntryImpl @Inject constructor(
                 notificationRenderer = notificationRenderer,
                 navController = childNavController
             )
+            ComposableSelfUpdater()
```
LionZXY, line 56: Why you check it in `Composable`?
You can move it to `Application` and init on each startup

components/bottombar/impl/src/main/java/com/flipperdevices/bottombar/impl/api/BottomNavigationFeatureEntryImpl.kt
```diff
@@ -49,6 +53,18 @@ class BottomNavigationFeatureEntryImpl @Inject constructor(
...
+        }
+    }
+
+    @Composable
+    private fun ComposableSelfUpdater() {
+        DisposableEffect(key1 = Unit) {
+            selfUpdaterApi.startCheckUpdate()
+
+            onDispose {
+                selfUpdaterApi.stopProcessCheckUpdate()
```
LionZXY, line 66: Why you unsubscribe on dispose? What happens if we don't do it?
I suggest don't call unsubscribe, what you think about it?

components/selfupdater/api/src/main/kotlin/com/flipperdevices/selfupdater/api/SelfUpdaterApi.kt
```diff
@@ -0,0 +1,6 @@
+package com.flipperdevices.selfupdater.api
+
+interface SelfUpdaterApi {
+    fun startCheckUpdate()
```
LionZXY, line 4: NIT: Let's add `...Async` to `startCheckUpdate` method: `startCheckUpdateAsync`

components/selfupdater/source/googleplay/src/main/kotlin/com/flipperdevices/selfupdater/googleplay/api/SelfUpdaterGooglePlayApi.kt
```diff
@@ -0,0 +1,65 @@
...
+private const val UPDATE_CODE = 228
+private const val NOTIFICATION_MS = 5000L
+
+@ContributesBinding(AppGraph::class, SelfUpdaterApi::class)
+class SelfUpdaterGooglePlayApi @Inject constructor(
+    private val context: Context,
+    private val inAppNotificationStorage: InAppNotificationStorage
+) : SelfUpdaterApi {
+
+    private val appUpdateManager = AppUpdateManagerFactory.create(context)
```
LionZXY, line 28: NIT: looks like heavy operation, so move it in `lazy` block

components/selfupdater/source/googleplay/build.gradle.kts
LionZXY, line 1: NIT: Move both package to `:selfupdater` level to decrease complexity

components/selfupdater/source/googleplay/src/main/kotlin/com/flipperdevices/selfupdater/googleplay/api/SelfUpdaterGooglePlayApi.kt
```diff
@@ -0,0 +1,65 @@
...
+    private val appUpdateInfoTask = appUpdateManager.appUpdateInfo
+    private var updateListener = InstallStateUpdatedListener(::processUpdateListener)
+
+    private fun processUpdateListener(state: InstallState) {
+        if (state.installStatus() != InstallStatus.DOWNLOADED) return
```
LionZXY, line 33: NIT: Add logs here, please

LionZXY (approved):

PR #561, opened 2023-04-05
Use release build for release

LionZXY:
**Background**

Right now we use internal for release build

**Changes**

- Use release build
- In RC build release build, but not publish it

**Test plan**

Try build release build


Programistich (approved):
Thx <3

PR #565, opened 2023-04-07
FDroid selfupdate

Programistich:
**Background**

Now we can self update if app install from fdroid

**Changes**

* Refactor modules like
- api
- googleplay
- noop
- thirdpart
   - api
   - github
   - fdroid

* Support fdroid check new release
* thirdpart module for get new version, compare, download

**Test plan**

Wait for new version and try update


instances/app/src/main/java/com/flipperdevices/app/FlipperApplication.kt
```diff
@@ -25,7 +25,7 @@ class FlipperApplication : Application() {
                 application = this,
                 ApplicationParams(
                     startApplicationClass = SingleActivity::class,
-                    version = BuildConfig.VERSION_NAME
+                    version = "1.3.9"
```
LionZXY, line 28: Please, return back

LionZXY (approved):

PR #569, opened 2023-04-11
Disable verbose ble logging by default

LionZXY:
**Background**

Right now ble have more flood in console

**Changes**

Disable ble log by default

**Test plan**

Try open application and see on logs


Programistich (approved):
Good WORK!

PR #574, opened 2023-04-18
Replace radio type to radio sub number in device info

Programistich:
**Background**

Replace radio type to radio type number from device info

**Changes**

For example, in previous variant we have Major.Minor.Type(Type Name)
And now we have Major.Minor.Sub(Type Name)

**Test plan**

Try open device info and see radio stack in Major.Minor.Sub(Type Name) format


LionZXY (approved):
Thank you sooo much!

PR #576, opened 2023-04-30
[Draft] Infrared key emulate

Programistich:
**Background**

**Changes**

Here what change in this PR..

**Test plan**

..and how we can test it


components/archive/shared/src/test/kotlin/com/flipperdevices/archive/shared/ExtractKeyMetaInformationTest.kt
```diff
@@ -37,7 +37,8 @@ class ExtractKeyMetaInformationTest(
                 FlipperKeyParsed.Infrared(
                     keyName = "Infrared",
                     notes = null,
-                    protocol = "TV"
+                    protocol = "TV",
```
Programistich, line 40: Test Comment for Github

PR #578, opened 2023-05-05
Unleashed

IhimeraBishop:
Darkflipper**Background**

**Changes**

Here what change in this PR..

**Test plan**

..and how we can test it


LionZXY:
Very cool!
There are a few things that could make your job easier:
1) You can simply turn off the metric with the [flag](https://github.com/flipperdevices/Flipper-Android-App/blob/dev/build-logic/plugins/convention/src/main/kotlin/com/flipperdevices/buildlogic/ApkConfig.kt#L36). Not a single line of code with the metric will get into your build, you can check this with [jadx](https://github.com/skylot/jadx), for example
2) Instead of skipping subghz provisioning you could provide the file you want, that way the app would update the file even on official firmware
3) Instead of hardcode replacing the update link you could make a switch in the settings so that even those using official firmware could use the app (like here: #452)
4) You can add support for self-updater. See [here](https://github.com/flipperdevices/Flipper-Android-App/blob/dev/components/selfupdater/thirdparty/github/src/main/kotlin/com/flipperdevices/selfupdater/thirdparty/github/parser/GithubParser.kt#L14)

Denlo94 (commented):
Nice

Denlo94 (commented):
README.md

PR #579, opened 2023-05-10
Add retry dialog on first pair

LionZXY:
**Background**

Right now we just stop pairing if connection failed

**Changes**

Suggest retry connection

**Test plan**

Try connect to flipper and drop/cancel connection

### DIFF
--- a/components/firstpair/impl/src/main/java/com/flipperdevices/firstpair/impl/composable/searching/ComposableSearchingDevices.kt
+++ b/components/firstpair/impl/src/main/java/com/flipperdevices/firstpair/impl/composable/searching/ComposableSearchingDevices.kt
@@ -1,15 +1,21 @@
 package com.flipperdevices.firstpair.impl.composable.searching
 
+import androidx.annotation.StringRes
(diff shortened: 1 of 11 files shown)


components/firstpair/impl/src/main/java/com/flipperdevices/firstpair/impl/composable/searching/ComposableSearchingDevices.kt
```diff
@@ -1,29 +1,58 @@
...
-    modifier: Modifier = Modifier,
-    onRefreshSearching: () -> Unit
+    onRefreshSearching: () -> Unit,
+    onResetTimeoutState: () -> Unit,
+    modifier: Modifier = Modifier
 ) {
     val devices = state.devices
 
-    SwipeRefresh(onRefreshSearching) {
+    if (state.pairState is DevicePairState.TimeoutPairing) {
```
Programistich, line 31: I think there we can use when

Programistich (changes requested):

Programistich (changes requested):
Some question

PR #620, opened 2023-06-16
Hide self updater dialog on click

Programistich:
**Background**

Now if u click on self update dialog, message wasn't hide

**Changes**

* Check Google Play build from gradle properties
* Self Updater for debug mode with flag
* Remove Open mfkey32 from options
* Hide message on click and disable action for handler in disposable effect
* Screen in bottom bar on full size

**Test plan**

* Wait for new version and try to click update


build-logic/plugins/convention/src/main/kotlin/com/flipperdevices/buildlogic/ApkConfig.kt
```diff
@@ -26,11 +26,16 @@ object ApkConfig {
         get() = prop("is_sentry_publish", false).toBoolean()
 
     val Project.SOURCE_INSTALL
-        get() = when (providers.gradleProperty("source_install").orNull) {
-            "fdroid" -> SourceInstall.FDROID
-            "github" -> SourceInstall.GITHUB
-            "googleplay" -> SourceInstall.GOOGLE_PLAY
-            else -> SourceInstall.UNKNOWN
+        get() = run {
+            if (VERSION_NAME == "DEBUG_VERSION") {
```
LionZXY, line 30: Please, move `"DEBUG_VERSION"` to constants

components/settings/impl/src/main/res/values/strings.xml
```diff
@@ -23,6 +23,7 @@
     <string name="debug_restart_rpc">Restart RPC</string>
     <string name="debug_open_mfkey32">Open MFKey32</string>
```
LionZXY, line 24: Please, remove it too

components/inappnotification/impl/src/main/java/com/flipperdevices/inappnotification/impl/composable/ComposableInAppNotification.kt
```diff
@@ -76,10 +80,10 @@ private fun ComposableInAppNotificationCard(
         val handler = Handler(Looper.getMainLooper())
 
         val fadeOutRunnable = {
-            visibleState = false
+            if (visibleState) visibleState = false
         }
         val hiddenRunnable = {
-            onNotificationHidden()
+            if (!actionClicked) onNotificationHidden()
```
LionZXY, line 86: How it works? Why you think that update dialog hide?

LionZXY (approved):

LionZXY (approved):

PR #629, opened 2023-06-25
Stream for crypto and share key content

Programistich:
**Background**

Now we save key content to ram, and this would lead to out of memory

**Changes**

* Crypto and send to server use streams
* Write Tests for encrypt/decrypt streams 

**Test plan**

Try share key and receive in variants Android/iOS


components/share/cryptostorage/src/main/kotlin/com/flipperdevices/share/cryptostorage/CryptoStorageApiImpl.kt
```diff
@@ -1,40 +1,98 @@
...
+            val decryptHelper = DecryptHelper()
+
+            val response = client.get(urlString = "${STORAGE_URL}$id/$STORAGE_NAME")
+
+            when (response.status) {
+                HttpStatusCode.OK -> {}
+                HttpStatusCode.NotFound -> throw FileNotFoundException()
+                else -> throw UnknownServiceException()
+            }
+            val inputStream = response.bodyAsChannel()
```
LionZXY, line 92: Use `inputStream` in `use` block for clean up after stream is finish

components/share/cryptostorage/src/main/kotlin/com/flipperdevices/share/cryptostorage/helper/DecryptHelper.kt
```diff
@@ -0,0 +1,66 @@
...
+                val cipherText = bytes.copyOfRange(IV_LENGTH, bytes.size)
+                cipher = Cipher.getInstance(ALGORITHM_HELPER).apply {
+                    init(
+                        Cipher.DECRYPT_MODE,
+                        secretKey,
+                        GCMParameterSpec(TAG_LENGTH * BIT_SIZE, iv)
+                    )
+                }
+                val decryptedBytes = cipher.doFinal(cipherText)
+                withContext(Dispatchers.IO) {
```
LionZXY, line 55: Maybe execute all method in IO context?

components/share/cryptostorage/src/main/kotlin/com/flipperdevices/share/cryptostorage/helper/EncryptHelper.kt
```diff
@@ -0,0 +1,66 @@
...
+        init(
+            Cipher.ENCRYPT_MODE,
+            secretKey,
+            GCMParameterSpec(TAG_LENGTH * BIT_SIZE, generateIV())
+        )
+    }
+
+    suspend fun writeEncrypt(outputChannel: ByteWriteChannel) {
+        val encryptBuffer = ByteArray(DEFAULT_BUFFER_SIZE)
+        val inputStream = flipperKeyContent.openStream()
```
LionZXY, line 36: You should write `use` here, just after create stream, because if something happens before `use` you will have a stream leak

LionZXY (approved):

PR #644, opened 2023-07-07
New Infrared Screen with emulate

Programistich:
**Background**

Now we cannot emulate infrared keys

**Changes**

* core:tabswitch for reuse in catalog and infrared screen
* Infrared Feature Entry (if key deleted, open default KeyScreen)
* Simple ShareBottomApi for use share key

**Test plan**

Try open infrared screen and press/hold/press diff button for emulate
Also try use actions in right top


sync-dev.sh
```diff
@@ -0,0 +1,9 @@
+current_branch=$(git rev-parse --abbrev-ref HEAD)
+
+git fetch upstream
+git checkout dev
+git merge upstream/dev
+
+git push
+
+git checkout $current_branch
```
LionZXY, line 9: Please, remove it

PR #647, opened 2023-07-10
Update strings.xml

Programistich:
**Background**

Lexical error on text

**Changes**

Add "L"

**Test plan**

Try run app


Programistich:
Fixed

PR #672, opened 2023-08-03
Add apk artifacts to Pull Requests

LionZXY:
**Background**

Right now we should download and build source code for testing PR

**Changes**

- Build apks for each PR

**Test plan**

See below for apks


Programistich (approved):
NIT: Use github secrets instead raw text in pr.yml

PR #683, opened 2023-08-17
Not send empty report in faphub

LionZXY:
**Background**

Right now we can send empty report from faphub

**Changes**

- Update report app icon
- Change Report App title

**Test plan**

Try report empty description


components/faphub/report/impl/src/main/kotlin/com/flipperdevices/faphub/report/impl/composable/ComposableReadyToReport.kt
```diff
@@ -60,7 +60,18 @@ internal fun ComposableReadyToReport(
...
+                submit(text)
+            }
+        },
+        enabled = validateReportText(text)
     )
 }
+
+private const val REPORT_TEXT_SIZE_MIN = 5
+
+private fun validateReportText(text: String): Boolean {
```
Programistich, line 74: NIT: Business logic check in View Models

Programistich (approved):

Programistich (approved):

PR #690, opened 2023-08-31
Refactor bottom sheet

Programistich:
**Background**

Now we use bottom sheet by navigation

**Changes**

* BottomBar setup color
* Open bottom sheet by action
* Colors by state sheet
* Start/Stop process sharing on subscribe sheet state

**Test plan**

Try share key


components/share/uploader/src/main/kotlin/com/flipperdevices/uploader/api/ShareBottomUIImpl.kt
```diff
@@ -1,66 +1,114 @@
...
         }
 
-        val navController = rememberNavController()
-        val bottomSheetNavigator = rememberBottomSheetNavigator()
-        navController.navigatorProvider += bottomSheetNavigator
+        val sheetState = rememberModalBottomSheetState(
+            initialValue = ModalBottomSheetValue.Hidden,
+            skipHalfExpanded = true
+        )
+        key(sheetState) {
```
LionZXY, line 50: NIT: Maybe change to `sheetState.isVisible`?

components/share/uploader/src/main/kotlin/com/flipperdevices/uploader/viewmodel/UploaderViewModel.kt
```diff
@@ -31,6 +30,7 @@ import java.net.UnknownHostException
 import java.net.UnknownServiceException
 
 private const val SHORT_LINK_SIZE = 256
+private const val EXTRA_KEY_PATH = "flipper_key_path"
```
LionZXY, line 33: In this place you should explicitly throw flipperKeyPath, because it is not known from which navigation scope can call this method.

LionZXY (approved):

PR #709, opened 2023-09-26
Flipper Busy Dialog with remote control

Programistich:
**Background**

Now on Flipper Busy Dialog we only information user

**Changes**

* New Flipper Busy Dialog + resources

**Test plan**

Open app in Flipper and try emulate key from phone


components/keyemulate/impl/src/main/java/com/flipperdevices/keyemulate/viewmodel/EmulateViewModel.kt
```diff
@@ -180,6 +183,10 @@ abstract class EmulateViewModel(
         }.launchIn(viewModelScope)
     }
 
+    fun goToRemoteScreen(navController: NavController) {
+        navController.navigate(screenStreamingFeatureEntry.ROUTE.name)
```
LionZXY, line 187: Please, use DI and `start` method

LionZXY (approved):

PR #713, opened 2023-10-03
Fix overflow in key small

Programistich:
**Background**

Now we can see more than 1 line on key small card

**Changes**

Setup max lines

**Test plan**

Choose big font size


LionZXY (approved):
Thank you!

PR #723, opened 2023-10-23
Fix crash on appication from unknown source

Programistich:
**Background**

Now we have crash on install

**Changes**

Return no updates on unknown source

**Test plan**

Try to install app from unknown source

### DIFF
--- a/components/selfupdater/impl/src/main/kotlin/com/flipperdevices/selfupdater/impl/api/SelfUpdaterApiImpl.kt
+++ b/components/selfupdater/impl/src/main/kotlin/com/flipperdevices/selfupdater/impl/api/SelfUpdaterApiImpl.kt
@@ -29,6 +29,9 @@ class SelfUpdaterApiImpl @Inject constructor(
         }
         return try {
             selfUpdaterSourceApi.checkUpdate(manual = manual)
+        } catch (e: Exception) {
+            info { "Self update error: $e" }
+            return SelfUpdateResult.ERROR
         } finally {
             inProgressState.emit(false)
         }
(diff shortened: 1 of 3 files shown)


components/selfupdater/unknown/src/main/kotlin/com/flipperdevices/selfupdater/unknown/api/SelfUpdaterUnknown.kt
```diff
@@ -9,7 +9,7 @@ import javax.inject.Inject
 @ContributesBinding(AppGraph::class, SelfUpdaterSourceApi::class)
 class SelfUpdaterUnknown @Inject constructor() : SelfUpdaterSourceApi {
     override suspend fun checkUpdate(manual: Boolean): SelfUpdateResult {
-        throw NotImplementedError()
```
LionZXY, line 12: What about try catch in `SelfUpdaterApiImpl#startCheckUpdate`?

Programistich: Make sense

components/selfupdater/impl/src/main/kotlin/com/flipperdevices/selfupdater/impl/api/SelfUpdaterApiImpl.kt
```diff
@@ -29,6 +29,9 @@ class SelfUpdaterApiImpl @Inject constructor(
         }
         return try {
             selfUpdaterSourceApi.checkUpdate(manual = manual)
+        } catch (e: Exception) {
+            info { "Self update error: $e" }
```
LionZXY, line 33: NIT: In this case you should use `error` logger

LionZXY (approved):

PR #731, opened 2023-11-01
New emulate api for infrared

Programistich:
**Background**

Now we can true emulate infrared with same name

**Changes**

* Update protobuf
* Check support new api
* New field in emulate config

**Test plan**

Try emulate with flipper on new api


LionZXY (approved):
Thank you for amazing work!

PR #734, opened 2023-11-09
Handle expired flipper crypto link

Programistich:
**Background**

Now we not correct handle exception after open expired link

**Changes**

* Use ktor exception ClientRequestException with check code

**Test plan**

* Try open https://flpr.app/sf#path=infrared/Selo_tv.ir&id=nLoer3&key=S0BWADnRlnhxQEU9jMwFsQ


LionZXY (approved):
Thank you for handle 404 error correctly <3

PR #740, opened 2023-11-15
Progress wrapper division by zero

### LINKED ISSUE #739: App crash when i tap Mfkey32 (Detect Reader)
**Describe the bug**
When I tap to "Mfkey32 (Detect Reader)", the application crashes.

Flipper_app

**To Reproduce**
- ".mfkey32.log" file is an empty file inside NFC folder.
- Open Flipper Android Application, sync the FlipperZero device via bluetooth, then Hub > NFC Tools > Mfkey32 (Detect Reader).
- App crashes.

**Target**
No app crashes.

**Additional context**
I deleted the ".mfkey32.log" file from Flipper Zero, then no other crashes when I tap "Mfkey32 (Detect Reader)"

Programistich:
**Background**

Now we have crash when file size equals

**Changes**

* Check divisions by zero, so set max percentage

**Test plan**

* Try use .mfkey32.log size equals zero

Fix #739


LionZXY (approved):
Thank you!

PR #749, opened 2023-12-12
Migrate options screen to decompose

LionZXY:
**Background**

Right now we have outdated navigation with google

**Changes**

Migrate options screen to decompose

**Test plan**

Try all screens in options category


components/analytics/shake2report/impl/src/main/java/com/flipperdevices/analytics/shake2report/impl/api/Shake2ReportDecomposeComponentImpl.kt
```diff
@@ -0,0 +1,38 @@
...
+
+class Shake2ReportDecomposeComponentImpl @AssistedInject constructor(
+    @Assisted private val onBack: DecomposeOnBackParameter,
+    private val viewModelProvider: Provider<Shake2ReportViewModel>
+) : Shake2ReportDecomposeComponent {
+    @Composable
+    @Suppress("NonSkippableComposable")
+    override fun Render() {
+        Shake2ReportScreen(
+            onBack = { onBack() },
```
Programistich, line 24: We can use lambda without {}

Programistich (changes requested):

PR #772, opened 2024-02-06
dependabot.yml

AbdhilahiRWabwire:
# Dependabot

- dependabot.yml - Automated dependency updates built into GitHub
 
- https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
 
- https://github.com/dependabot

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -16,47 +16,11 @@ updates:
       day: "saturday"
       interval: "weekly"
       time: "03:00"
-
-  - package-ecosystem: "gradle"
-    commit-message:
-      prefix: "Gradle"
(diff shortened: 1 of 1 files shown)


LionZXY:
It looks like you're looking in gradle scripts to extract dependencies, but that's wrong. 
The project uses the [Gradle Version Catalog](https://docs.gradle.org/current/userguide/platforms.html) and you should check for updates to it

AbdhilahiRWabwire:
@LionZXY Thank you for that information. I fixed the file.

LionZXY:
Thank you for your contribution!

PR #798, opened 2024-03-22
Use anvil-utils to generate and contribute assisted factories bindings

IlyaGulya:
**Background**

I've noticed that there's a lot of boilerplate assisted factories in the project.
Recently I've created a code generator for anvil which allows to get rid of this boilerplate. (https://github.com/IlyaGulya/anvil-utils)

**Changes**

This PR replaces all Assisted factories bindings with `@ContributesAssistedFactory` annotation which automatically generates assisted factory with all `@Assisted` constructor arguments and binds it to provided scope and public API.

**Test plan**

Thanks to Dagger, if it compiles, it works 🙂


build-logic/plugins/convention/src/main/kotlin/flipper.anvil.gradle.kts
```diff
@@ -10,4 +10,6 @@ anvil {
 
 dependencies {
     "implementation"(libs.dagger)
+    "api"(libs.anvil.utils.annotations)
```
LionZXY, line 13: Why `api`? Maybe `implementation` is more useful here?

IlyaGulya: That's because `flipper.anvil` plugin is not applied to `:instances:*` and therefore `libs.anvil.utils.annotations` is not on the classpath of anvil code generator. (I will investigate it and maybe remove this requirement in the future version of code generator)
For now, to be able to use `implementation` we should do something of this:
- apply `flipper.anvil` plugin to `:instances:*`
- add `implementation(libs.anvil.utils.annotations)` to all `:instances:*` modules

IlyaGulya: Oh, it's not even my code generator issue, but general issue that kapt can't work with code which contains something which is not on the classpath of module where it's called.
So, there's only one of the options above is applicable, nothing to do with my code generator 🙂

IlyaGulya: I think the best way would be to apply `flipper.anvil` to instances.
If you're worried about anvil complaints regarding `generateDaggerFactories` being enabled in the module where Dagger is also applied, I have such workaround in my project, we can apply it here as well:

```
plugins {
    id("com.squareup.anvil")
}

anvil {
    generateDaggerFactories.set(true)
}

pluginManager.withPlugin(libs.plugins.kotlin.kapt.get().pluginId) {
    anvil {
        generateDaggerFactories.set(false)
    }
}
```
This way `generateDaggerFactories` will become false if `kapt` is applied to module. (Dagger is the only `kapt` dependency in my project, so this heuristic works fine)

IlyaGulya: Let me know which option do you like more and I will fix that 🙂

LionZXY:
Again - thank you so much for the great work you do

LionZXY (approved):

PR #806, opened 2024-04-03
Find paired flipper devices without state filter

MatkovIvan:
**Background**

In my environment (Nothing Phone (2), Android 14), the app cannot find flipper device via Bluetooth. During debugging I found that scanning callbacks not called at all, but in the same time I can easily found & pair device via Bluetooth settings. After pairing, the app is still unable to find it even in connected devices.

**Changes**

- Use `bluetoothAdapter.bondedDevices` instead of `bluetoothManager.getConnectedDevices` during searching that filters devices by `BluetoothProfile.STATE_CONNECTED`
- Remove `getAlreadyBoundedDevice(deviceId: String)` as it's already checked in the same list
- Fix typo (bounded -> bonded)

Please note that the change is made with assumption that `bluetoothManager.getConnectedDevices` returns subset of `bluetoothAdapter.bondedDevices`. It's true for my case, but needs to be re-checked.

**Test plan**

Pair the device via Bluetooth settings and try to find it via app after that.


LionZXY:
I have checked this in all configurations. Everything works much better than it was. Thanks a lot for your contribution!

Can I ask you to add a line to the Changelog and correct the tests?

LionZXY (approved):

PR #807, opened 2024-04-04
Fix/infrared remote placeholders design

makeevrserg:
**Background**

Previously flipper android app had two ui states `ErrorContent` and `ComposableInfraredRemotes`. This PR intends to add more UI states which will help user to understand current `EmulatorState` more precisely

**Changes**

- Add fade modifier
- Fix placeholder color
- Fix edit remote save button color
- Add syncing ui state display
- Add empty remote buttons hint
- Improve multiple buttons display upon flipper not connected 

**Test plan**

- Open infrared key info
- Disconnect flipper device to see disconnected state with no remote buttons
- Disconnect flipper device to see disconnected state with 1-2 remote buttons
- Disconnect flipper device to see disconnected state with >3 remote buttons
- Emulate low connection to see connecting state
- Connect flipper to see connected state with remote buttons
- Connect flipper to see connected state without remote buttons


components/core/ui/ktx/src/main/java/com/flipperdevices/core/ui/ktx/fade/FadeKtx.kt
```diff
@@ -0,0 +1,85 @@
...
+import androidx.compose.ui.graphics.BlendMode
+import androidx.compose.ui.graphics.Brush
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.CompositingStrategy
+import androidx.compose.ui.graphics.graphicsLayer
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import com.flipperdevices.core.ui.theme.FlipperThemeInternal
+
+fun Modifier.fadingEdge(orientation: FadeOrientation): Modifier {
```
makeevrserg, line 19: This is common design patter, so it's probably better to have in core-ui with `FadeOrientation`

components/keyemulate/impl/src/main/java/com/flipperdevices/keyemulate/api/KeyEmulateUiApiImpl.kt
```diff
@@ -58,4 +59,13 @@ class KeyEmulateUiApiImpl @Inject constructor() : KeyEmulateUiApi {
             progressColor = progressColor
         )
     }
+
+    @Composable
+    override fun ComposableEmulatePlaceholderButton(
+        modifier: Modifier
+    ) {
+        ComposableEmulatePlaceholder(
```
makeevrserg, line 67: Pretty Doubtful to make something like this, but Composables of it's modules are inside impl, so this is probably the only right way to do it

components/infrared/impl/src/main/kotlin/com/flipperdevices/infrared/impl/composable/components/ComposableIconText.kt
```diff
@@ -0,0 +1,72 @@
...
+                    .padding(horizontal = 4.dp)
+                    .size(14.dp),
+                painter = painter,
+                contentDescription = text,
+                tint = LocalPallet.current.warningColor
+            )
+        }
+        Text(
+            text = text,
+            style = LocalTypography.current.bodyR14.copy(fontSize = 12.sp),
```
LionZXY, line 57: I realise it's not obvious now, but in this case it's better to use `subtitleR12` or create a new font. 
I realise that this is not the best solution, but I would like to get rid of manually setting the font size in the project

components/infrared/impl/src/main/kotlin/com/flipperdevices/infrared/impl/composable/components/ComposableInfraredRemotes.kt
```diff
@@ -17,11 +27,68 @@ import kotlinx.collections.immutable.toImmutableList
...
+) {
+    val remoteConfigs = remember(state) {
+        val allConfigs = state.toEmulateConfigs()
+        allConfigs.subList(0, FADE_LIST_THRESHOLD.coerceAtMost(allConfigs.size))
+    }
+
+    InfraredRemotesContent(
+        remoteConfigs = remoteConfigs,
+        keyEmulateContent = keyEmulateContent,
+        fadeOrientation = FadeOrientation.Bottom()
```
LionZXY, line 59: It looks as if a new object will be created with each recomposition. Is there any way to fix this?

components/keyemulate/impl/src/main/java/com/flipperdevices/keyemulate/api/KeyEmulateUiApiImpl.kt
LionZXY: Looks good, thank you for highlight this case

components/infrared/impl/src/main/kotlin/com/flipperdevices/infrared/impl/composable/components/ComposableInfraredRemotes.kt
makeevrserg: Either we can add Compose's `remember {}` or use `FadeOrientation.Bottom` as object, but changing threshold of fade will not be possible in this case. remember should be enough to fix the case of recomposition

components/core/ui/ktx/src/main/java/com/flipperdevices/core/ui/ktx/fade/FadeOrientation.kt
```diff
@@ -0,0 +1,19 @@
...
+    ) : FadeOrientation
+
+    class Top(
+        @FloatRange(0.0, 1.0)
+        val threshold: Float = DEFAULT_FADE_THRESHOLD
+    ) : FadeOrientation
+
+    companion object {
+        private const val DEFAULT_FADE_THRESHOLD = 0.7f
+    }
```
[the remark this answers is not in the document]
makeevrserg: Replaced with data object as for now no need in changing threshold

components/core/ui/ktx/src/main/java/com/flipperdevices/core/ui/ktx/fade/FadeKtx.kt
```diff
@@ -0,0 +1,85 @@
...
+    val fillColor = Color.White
+    val brush = orientation.toBrush(fillColor)
```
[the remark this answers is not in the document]
makeevrserg: Added `remember` to this extension

components/infrared/impl/src/main/kotlin/com/flipperdevices/infrared/impl/composable/components/ComposableIconText.kt
makeevrserg: `subtitleM12` actually fits perfectly here, fixed

LionZXY (approved):
LGTM

PR #808, opened 2024-04-04
Fix/no internet connection on hub apps

makeevrserg:
Background

On Hub/Apps screen displayed multiple error placeholders when categories and list could not be loaded. There's should be either displayed categories with list error or fullscreen error in case categories could not be loaded

Changes

- Removed entirely error placeholder from ComposableCategories as it's not required now
- Block list error display when categories error happened
- Add fullscreen LazyColumn error display

Test plan

- Get Charles or HTTP Toolkit to simultate url timeouts or just disable internet on phone
- Open Hub>Apps
- Await categories loaded and simulate error for list below to see it's error placeholder
- Restart app with no internet to see fullscreen error placeholder on this screen when categories not loaded

### DIFF
--- a/components/faphub/catalogtab/impl/src/main/java/com/flipperdevices/faphub/catalogtab/impl/composable/categories/ComposableFullScreenError.kt
+++ b/components/faphub/catalogtab/impl/src/main/java/com/flipperdevices/faphub/catalogtab/impl/composable/categories/ComposableFullScreenError.kt
@@ -2,29 +2,23 @@ package com.flipperdevices.faphub.catalogtab.impl.composable.categories
 
 import androidx.compose.foundation.lazy.LazyListScope
 import androidx.compose.ui.Modifier
-import com.flipperdevices.faphub.catalogtab.impl.model.CategoriesLoadState
 import com.flipperdevices.faphub.errors.api.FapErrorSize
 import com.flipperdevices.faphub.errors.api.FapHubComposableErrorsRenderer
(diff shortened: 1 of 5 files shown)


components/faphub/catalogtab/impl/src/main/java/com/flipperdevices/faphub/catalogtab/impl/composable/categories/ComposableFullScreenError.kt
```diff
@@ -0,0 +1,30 @@
...
+@Suppress("FunctionNaming")
+internal fun LazyListScope.ComposableFullScreenError(
+    categoriesLoadState: CategoriesLoadState,
+    errorsRenderer: FapHubComposableErrorsRenderer,
+    onRetry: () -> Unit,
+    modifier: Modifier = Modifier
+) {
+    (categoriesLoadState as? CategoriesLoadState.Error)
+        ?.throwable
+        ?.toFapHubError()
```
LionZXY, line 19: I think 
```
 (categoriesLoadState as? CategoriesLoadState.Error)
        ?.throwable
        ?.toFapHubError()
```
should be in `remember`

makeevrserg: Fixed `remember` and removed overhead

LionZXY (approved):
Thank you for this elegant solution!

Issue #737, opened 2023-11-13
WearOS interface very narrow


noiob:
Attempting to send any of the saved signals from my Pixel Watch yields a very narrow interface that doesn't accept any touch inputs, so sending doesn't work. Scrolling works. The archive list is fine.

![IMG_20231113_155658_132](https://github.com/flipperdevices/Flipper-Android-App/assets/8197071/c5a2c144-ec1a-4a6b-8223-9c0706203c09)
![IMG_20231113_155700_747](https://github.com/flipperdevices/Flipper-Android-App/assets/8197071/d30d2d2c-5490-49db-af6f-2315213bc5ca)

**Target**
version 1.6.6.1303
WearOS 4.0 on a Google Pixel Watch 1

PR #822, opened 2024-04-15
Fix wear OS narrow interface on button dispatch

makeevrserg:
**Background**

Currently UI layout on buttons dispatch screen is [very narrow](https://github.com/flipperdevices/Flipper-Android-App/issues/737)

**Changes**

- Fixed doubled paddings on buttons dispatch screen

**Test plan**

- Open wear os app
- Open screen with buttons dispatch
- See the size of buttons now looks good and not narrow

### DIFF
--- a/components/wearable/emulate/wear/impl/src/main/java/com/flipperdevices/wearable/emulate/impl/composable/ComposableWearEmulate.kt
+++ b/components/wearable/emulate/wear/impl/src/main/java/com/flipperdevices/wearable/emulate/impl/composable/ComposableWearEmulate.kt
@@ -1,6 +1,7 @@
 package com.flipperdevices.wearable.emulate.impl.composable
 
 import androidx.compose.foundation.layout.Box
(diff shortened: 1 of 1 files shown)


components/wearable/emulate/wear/impl/src/main/java/com/flipperdevices/wearable/emulate/impl/composable/ComposableWearEmulate.kt
```diff
@@ -24,7 +24,7 @@ fun ComposableWearEmulate(
 ) {
     val state by emulateViewModel.getWearEmulateState().collectAsState()
     val modifier = Modifier
-        .fillMaxRectangle()
+        .fillMaxWidth()
```
makeevrserg, line 27: `com.flipperdevices.wearable.core.ui.components.ComposableWearOsScrollableColumn` column already have by default `fillMaxRectangle` and it was also applied here. So here's required `fillMaxWidth` instead

PR #827, opened 2024-04-18
Replace Dispatchers.Default

makeevrserg:
**Background**

Right now, `Dispatchers.Default` is used in whole application, including `viewModelScope`. As it turns out, jvm Dispatcher created with `newWorkStealingPool` provides better perfomance than `Dispatchers.Default` itself.

**Changes**

- Replaced `Dispatchers.Default` with `FlipperDispatchers.workStealingDispatcher`

**Test plan**

- This PR requires to check every aspect of the application 😢 , but there's shouldn't be visible changes


LionZXY (approved):
Thank you for deep investigation of this bug!

PR #828, opened 2024-04-18
Fix/update turned off flipper wrong message

makeevrserg:
**Background**

Mobile application display wrong error message when disable flipper during update phase

**Changes**

- Fix when cases of `FAILED_SUB_GHZ_PROVISIONING`

**Test plan**

- Open mobile application
- Waith for flipper to sync
- Select random update channel
- Click install
- After downloading/installing started immediatly start reloading/disabling flipper
- See that error now displayed correctly


LionZXY (approved):
Thank you! It's a huge problem which happens sometimes, but we can't debug it

PR #830, opened 2024-04-19
Fix low resolution of saved screenshots

makeevrserg:
**Background**

When sharing screenshot on ScreenshotPreview screen it looks like screenshot very blurry. It's actually not blured, but low-resolutioned.

**Changes**

- Created BitmapKtx where rescale operation of bitmap is shared

**Test plan**

- Open flipper apps screenshots screen
- Share a screenshot 
- See screenshot now doesn't looks blurry


LionZXY (approved):
Thank you!

PR #835, opened 2024-04-24
Baseline profiles

makeevrserg:
**Background**

Right now after every release google collect profiles itself, but we can use baseline profiles so event the users who install updates first will see perfomance boost

**Changes**

- Added baseline profiles module for initial launch
- Updated CI to handle biseline files generation

**Test plan**
- Run CI on act
- Run CI on repository
- See statistic on GooglePlay Consolse


.github/workflows/calls/create-baseline-call.yml
```diff
@@ -0,0 +1,43 @@
+name: "detekt-validation"
+on:
+  workflow_call:
+    inputs:
+      BUILD_VARIANT_NAME:
+        required: true
```
makeevrserg, line 6: This custom workflow file allows to reuse it with different build variants

But didn't find a better way to use uppercase :(

LionZXY: Why not use bash for transform build variant name to uppercase?

.github/workflows/calls/create-baseline-call.yml
```diff
@@ -0,0 +1,43 @@
...
+        uses: actions/checkout@v2
+
+      - name: Set up JDK 17
+        uses: actions/setup-java@v2
+        with:
+          distribution: 'adopt'
+          java-version: 17
+
+      - name: Give gradle permission
+        run: chmod +x gradlew
```
LionZXY, line 26: Are you sure it's necessary?

.github/workflows/calls/create-baseline-call.yml
```diff
@@ -0,0 +1,43 @@
...
+  workflow_call:
+    inputs:
+      BUILD_VARIANT_NAME:
+        required: true
+        type: string
+      BUILD_VARIANT_NAME_UPPERCASE:
+        required: true
+        type: string
+jobs:
+  upload_baseline_file:
```
LionZXY, line 12: Let's use a different name for the job, shall we? How about the neutral ‘baseline_profile’? Now it may feel like it only loads baseline

.github/workflows/internal.yml
```diff
@@ -51,17 +51,28 @@ jobs:
...
         target: [ "gh_gms", "gh_nogms", "googleplay" ]
     steps:
       - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4
         with:
           submodules: 'recursive'
+      - uses: actions/download-artifact@master
+        name: "Download internal baseline file"
+        with:
+          name: baseline_profile_file_internal
+          path: ./instances/app/src/internal/generated/baselineProfiles/baseline-prof.txt
```
LionZXY, line 75: It seems wrong to me that job should know about the internal path in which the baseline profile is located. How about using outputs? And make a universal path - e.g. always save to ./baseline-prof.txt

components/baselineprofile/build.gradle.kts
LionZXY, line 1: I'm not sure `components/` is the right place for such a module. The concept is that components are the building bricks of an application - and here the baseline module clearly knows about `instances:app`, which violates the principles of modules in components. I think it has more space in `instances/` - as ideologically there are entry points into the application. What do you think?

instances/app/build.gradle.kts
```diff
@@ -263,4 +264,14 @@ dependencies {
     kapt(libs.dagger.kapt)
 
     implementation(libs.timber)
+
+    implementation(libs.profileinstaller)
+    "baselineProfile"(project(":components:baselineprofile"))
```
LionZXY, line 269: Is that how it's supposed to be or is that a typo?

components/baselineprofile/build.gradle.kts
makeevrserg: Yeah, it was initially inside `:instances` but I though `instances` only for application instances and `baselineprofile` is more like testing-only module. But it really knows about `instances:app`, so it's really doesn't belong here

instances/app/build.gradle.kts
makeevrserg: Quite a typo, forget to change from baselineorofile page

components/baselineprofile/build.gradle.kts
makeevrserg: Moved to `instances:android:baselineprofile`

.github/workflows/calls/create-baseline-call.yml
makeevrserg: F

.github/workflows/internal.yml
makeevrserg: We need to save it into specific folder, where it located. I created path for it to reuse later `create_baseline_file_path`

.github/workflows/calls/create-baseline-call.yml
makeevrserg: Yeah, the project's gradlew already has permission in project repo, not necessary

LionZXY (approved):
Looks perfect! Thank you!

PR #842, opened 2024-05-02
Fix md5 collision in DatabaseKeyContentConverter

makeevrserg:
**Background**

Currently, `DatabaseKeyContentConverter` doesn't check for same file content of generated md5 hash, which can lead to path collision

**Changes**

- Added `FileComparator` to compare two files/InputStreams
- Added `MD5Converter` to be able to mock md5 hash result
- Added tests as a proof of concept

**Test plan**

- Run tests
- Launch this feature in app


components/bridge/dao/impl/src/main/java/com/flipperdevices/bridge/dao/impl/di/RoomDatabaseModule.kt
```diff
@@ -22,19 +28,51 @@ private const val DATABASE_NAME = "flipper.db"
...
+    @Provides
+    fun provideFileComparator(): FileComparator {
+        return DefaultFileComparator
+    }
+
+    @Provides
+    fun provideMD5Converter(): MD5Converter {
+        return MD5ConverterImpl()
+    }
+
```
LionZXY, line 75: Why not use anvil ContributesBinding/Inject annotation for creating object without module? I think it will be better and clear, what you think about it?

components/bridge/dao/impl/src/main/java/com/flipperdevices/bridge/dao/impl/di/RoomDatabaseModule.kt
```diff
@@ -22,19 +28,51 @@ private const val DATABASE_NAME = "flipper.db"
...
+        context: Context,
+        fileComparator: FileComparator
+    ): MD5FileProvider = MD5FileProviderImpl(
+        context = context,
+        fileComparator = fileComparator
+    )
+
+    @Provides
+    fun provideFileComparator(): FileComparator {
+        return DefaultFileComparator
```
LionZXY, line 68: Maybe don't make the DefaultFileComparator object? It's strange that it will always live in memory. Maybe we should use scope @Reusable here?

components/bridge/dao/impl/src/main/java/com/flipperdevices/bridge/dao/impl/di/RoomDatabaseModule.kt
makeevrserg: Yes, replaced with injects

components/bridge/dao/impl/src/main/java/com/flipperdevices/bridge/dao/impl/di/RoomDatabaseModule.kt
makeevrserg: Fixed to class, added inject

components/bridge/dao/impl/src/test/java/com/flipperdevices/bridge/dao/impl/api/key/FileComparatorTest.kt
```diff
@@ -0,0 +1,113 @@
...
+import java.io.InputStream
+
+class FileComparatorTest {
+    private lateinit var fileComparator: FileComparator
+    private lateinit var tempFolder: File
+
+    @Before
+    fun setUp() {
+        fileComparator = DefaultFileComparator()
+        tempFolder = File(System.getProperty("java.io.tmpdir"), "TEMP_DIR")
```
LionZXY, line 21: Hm, it looks like this system api sometimes can not working. Let's see to TemporaryFolder rule in junit
https://junit.org/junit4/javadoc/4.13/org/junit/rules/TemporaryFolder.html

LionZXY (dismissed):
Thank you! It's super high quality PR!

LionZXY (approved):

PR #845, opened 2024-05-06
Fix filter dev catalog apps

makeevrserg:
**Background**

Right now flipper mobile app doesn't filter aps from prod/dev catalog. Because of this, after switching do dev catalog, apps from prod is displayed and vice versa

**Changes**
- Provided changes are the same as in [Flipper-iOS-App](https://github.com/flipperdevices/Flipper-iOS-App/commit/26553e7b7099ff349a631d39993571aa426c19d4)
- Added flag `isDevCatalog` to `FapManifestItem`
- Add filter inside `FapManifestsLoader` for `isDevCatalog`

> [!WARNING]  
> All installed apps which doesn't has a flag of `isDevCatalog` will return `isDevCatalog=false`

**Test plan**

1. [Optionally] delete all other apps for easier testing
2. Turn on usage of dev catalog
3. Install dev catalog apps
4. Turn off dev catalog
5. Open installed apps and see no dev-catalog apps present
6. Repeat steps 2-5 for prod-catalog

### DIFF
--- a/components/faphub/installation/manifest/impl/src/main/java/com/flipperdevices/faphub/installation/manifest/impl/utils/FapManifestParser.kt
+++ b/components/faphub/installation/manifest/impl/src/main/java/com/flipperdevices/faphub/installation/manifest/impl/utils/FapManifestParser.kt
@@ -18,6 +18,7 @@ private const val FAP_MANIFEST_PATH_KEY = "Path"
 private const val FAP_MANIFEST_FULL_NAME_KEY = "Full Name"
 private const val FAP_MANIFEST_ICON_KEY = "Icon"
 private const val FAP_MANIFEST_API_KEY = "Version Build API"
+private const val FAP_MANIFEST_DEV_CATALOG = "DevCatalog"
 
 class FapManifestParser @Inject constructor() {
     fun parse(fileContent: ByteArray, name: String): FapManifestItem? {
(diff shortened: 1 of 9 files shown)


components/faphub/installation/manifest/impl/src/main/java/com/flipperdevices/faphub/installation/manifest/impl/utils/FapManifestParser.kt
```diff
@@ -55,6 +57,8 @@ class FapManifestParser @Inject constructor() {
         orderedDict.add(FAP_MANIFEST_VERSION_UID_KEY to fapItem.versionUid)
         orderedDict.add(FAP_MANIFEST_PATH_KEY to fapItem.path)
 
+        orderedDict.add(FAP_MANIFEST_DEV_CATALOG to fapItem.isDevCatalog.toString())
```
LionZXY, line 60: I wouldn't want a debugging feature to affect the user experience in production. Increasing the size of the manifest may affect the speed of file loading - this is critical. Let's not add this field if its value is `false`?

makeevrserg: Makes sense. Added check for `fapItem.isDevCatalog`

LionZXY (approved):
Thank you!

PR #858, opened 2024-05-15
Fix sd card not mounted display

makeevrserg:
**Background**

On Hub->Installed sd card error displayed even if it's loaded. This happens because of long `cacheLoader.loadCache` operation

**Changes**

- Added initial state before `cacheLoader.loadCache` operation

**Test plan**

- Open Hub->Installed
- See SD Card error now does not display as long as before
- Additionaly add more logs ore debug breakpoint into `FapManifestsLoader` to see `NoSdCardException` message


LionZXY (approved):
Thank you for this quick fix)

PR #859, opened 2024-05-16
Fix/scroll of zoomed screenshotpreview

makeevrserg:
**Background**

When extremely fast scrolling zoomed image just after zoom, the image will be displayed above other image. It also can be reproduced by clicking on bottom image.

**Changes**

- Disable `HorizontalPager` zoom when image is zoomed
- Add ViewModel with events, which will be handled inside ViewPager. It will reset scale and only then swipe pages 

**Test plan**

- Open any screenshot preview screen
- Try zoom and extrmely fast scroll
- See it's now imposible
- Try zoom image and select any other image on bottom list
- See other image displayed without previous image overlay


components/faphub/screenshotspreview/impl/src/main/kotlin/com/flipperdevices/faphub/screenshotspreview/impl/composable/content/ComposableScreenshotsList.kt
```diff
@@ -80,11 +77,13 @@ internal fun ComposableScreenshotsList(
         horizontalArrangement = Arrangement.spacedBy(8.dp)
     ) {
         itemsIndexed(screenshots) { index, item ->
+            val isSelected = index == currentPage
```
makeevrserg, line 80: No `remember` required here as we need it updated every recomposition of `pagerState`

LionZXY (approved):

PR #862, opened 2024-05-22
Fix infrared keys list padding

makeevrserg:
**Background**

On infrared screen, when enabled android buttons gesture navigation, the last button located beyound bottom navigation

**Changes**

- Add element into list with bottom nav panel padding. The padding not added to column itself to save transparent display of navigation panel.

**Test plan**
- Enabled buttons panel navigation in device settings
- Open infrared editor with more keys than screen height
- Scroll down
- See the last element now not located beyound bottom navigation


LionZXY (approved):
Thank you for your attention to such small details :3

LionZXY (approved):

PR #863, opened 2024-05-23
Fix/markdown code block color

makeevrserg:
**Background**

Makrdown code blocks on flipper update screen has wrong colors

**Changes**

- Fixed markdown theme colors

**Test plan**

- Find somewhere in app markdown with code block(for example in flipper update to rc version)
- See that markdown code block now have readeable colors
- Also here's screenshot below
image


LionZXY (approved):
Thank you! And sorry for this bug

PR #868, opened 2024-05-27
Update to Kotlin 2.0

LionZXY:
**Background**

Kotlin 2.0 is the next generation of Kotlin language development. The transition to it should be smooth, but we are using compiler plugins and there may be problems because of that
More info: https://github.com/square/anvil/issues/733

**Changes**

- Update kotlin to 2.0
- Update anvil to 2.5.0-beta09
- Update anvil-utils
- Migrate to Kotlin Compose gradle plugin

**Test plan**

Try to build app


LionZXY:
Waiting for a fix: https://github.com/gradle/gradle/issues/27099

makeevrserg (approved):

PR #875, opened 2024-06-17
Design issues

makeevrserg:
**Background**

Tools tab contains NFCAttack, but only mfkey located here. Also the design of badges and connected tab button needs to be changed

**Changes**

- remove entirely nfcattack module
- move mfkey to tools tabs
- Fix badges outline colors
- Fix tab connection active button color

**Test plan**

- Open tools tab
- Open mfkey screen
- Check first tab for new colors


LionZXY (approved):
Thank you for fast fix!

PR #882, opened 2024-07-01
Workaround for Flipper connection mac change

LionZXY:
**Background**

Right now we have when reconnecting, the user has to guess that they should reconnect the flipper. 
We can on Android search by device name to tell the user that the MAC address of their device has changed

Related to: https://github.com/flipperdevices/flipperzero-firmware/pull/3723

**Changes**

- Now we use our local logic to reconnect instead of the system logic
- We search for the device by device name if the mac address connection fails

**Test plan**

Try change mac address and reconnect to device


makeevrserg (dismissed):

makeevrserg (approved):

Hayden925-ops (commented):
I need some help with understanding the flipper

fantomazz26 (commented):
Ссбцкифъ

WooZyCrAsH (commented):
Let's go!!

krokobrandt-boop (commented):
> - - [ ]

PR #891, opened 2024-07-15
Switch to self-hosted github runners

drunkbatya:
**Background**

Switch to self-hosted github runners

**Changes**

Switch to self-hosted github runners

**Test plan**

All PR should works

### DIFF
--- a/.github/workflows/internal.yml
+++ b/.github/workflows/internal.yml
@@ -11,7 +11,7 @@ concurrency:
 jobs:
   invalidate_gradle_cache:
     name: Update gradle cache
-    runs-on: ubuntu-latest
+    runs-on: [self-hosted, AndroidShell]
     steps:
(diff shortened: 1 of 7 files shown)


.github/workflows/internal.yml
```diff
@@ -11,36 +11,40 @@ concurrency:
 jobs:
   invalidate_gradle_cache:
     name: Update gradle cache
-    runs-on: ubuntu-latest
+    runs-on: [self-hosted, AndroidShell]
     steps:
-      - uses: actions/checkout@0ad4b8fadaa221de15dcec353f45205ec38ea70b # v4
+      - uses: actions/checkout@v4.1.1 # v4
```
LionZXY, line 16: FYI: We use a clear commit indication for a reason, but to exclude vulnerability with substitution of other code in the tag

drunkbatya: Ok, np

PR #892, opened 2024-07-15
Fix expandable signal verify

makeevrserg:
**Background**

When selecting signal inside Infrared remote setup - the "bottom sheet" displayed from the beginning. It should be displayed only when  signal is dispatched.

**Changes**

- Add state to display "bottom sheet" only after signal is dispatched

**Test plan**

- Open signal select screen
- Press signal
- See the confirm menu animated and shown only after signal is dispatched
- Press yes/no
- See "bottom sheet" is hidden again

### DIFF
--- a/components/remote-controls/setup/impl/src/main/kotlin/com/flipperdevices/remotecontrols/impl/setup/presentation/viewmodel/DispatchSignalViewModel.kt
+++ b/components/remote-controls/setup/impl/src/main/kotlin/com/flipperdevices/remotecontrols/impl/setup/presentation/viewmodel/DispatchSignalViewModel.kt
@@ -22,6 +22,7 @@ import com.flipperdevices.remotecontrols.impl.setup.util.toByteArray
 import com.squareup.anvil.annotations.ContributesBinding
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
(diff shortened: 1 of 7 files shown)


components/remote-controls/setup/impl/src/main/kotlin/com/flipperdevices/remotecontrols/impl/setup/presentation/viewmodel/DispatchSignalViewModel.kt
```diff
@@ -40,8 +41,20 @@ class DispatchSignalViewModel @Inject constructor(
...
     override val state = _state.asStateFlow()
+
+    private val _isEmulated = MutableStateFlow(false)
+    override val isEmulated = _isEmulated.asStateFlow()
+
     private var latestDispatchJob: Job? = null
 
+    override fun reset() {
+        viewModelScope.launch {
+            latestDispatchJob?.cancelAndJoin()
```
makeevrserg, line 52: May be here we should just join and not cancel latest job?

LionZXY: Cancel is fine

LionZXY (approved):

PR #961, opened 2024-09-23
Add better information during sync

makeevrserg:
**Background**

Current synchronization screen doesn't show us any details about current synchronization

**Changes**

- Edit ProgressTracker to add details into it
- Add more details to current syncrhonization

**Test plan**

- Open app
- See there's now more steps and you can see what now is actually syncing


LionZXY (approved):
It's really cool! Good job!

PR #981, opened 2024-10-30
Fix bytes race

makeevrserg:
**Background**

When sending a lot of bytes to flipper, some of them may be missed. This is happening because of a bug inside new `FlipperSerialOverflowThrottler`

**Changes**

- Fix receive for bytearray
- Fix long-polling request

**Test plan**

- Open sample (probably a 10 times to see that flipper now connecting normally)
- Open file manager and upload at least >1024KiB-sized file
- (or) open some key in text editor and save it as another file
- See uploaded/saved file has the same size as original


components/bridge/connection/feature/rpc/impl/src/commonMain/kotlin/com/flipperdevices/bridge/connection/feature/rpc/impl/FRpcFeatureApiImpl.kt
```diff
@@ -151,12 +156,23 @@ class FRpcFeatureApiImpl @AssistedInject constructor(
             }
         }.launchIn(scope + FlipperDispatchers.workStealingDispatcher)
 
+
+        val pingJob = scope.launch {
```
makeevrserg, line 160: Need to tell lags detector that we're fine when executing long-polling tasks. But may be there's should be different implementation

LionZXY (approved):

PR #985, opened 2024-11-08
Fix file upload progress

makeevrserg:
**Background**
This PR aims to improve new file manager uploading reactivity and fix ui for multiple files

**Changes**

- Remove upload screen and replace it with component
- Add callbacks for uploaded file progreess - will now be updated in items list
- Fix multiple files upload display 

**Test plan**

- Try upload one file fully and see it will appear in list instantly
- Try upload one file but cancel it in progress. See it also appeared, but with non-full upload size
- Try upload multiple files


components/filemngr/listing/impl/src/commonMain/kotlin/com/flipperdevices/filemanager/listing/impl/api/FilesDecomposeComponentImpl.kt
```diff
@@ -55,10 +57,21 @@ class FilesDecomposeComponentImpl @AssistedInject constructor(
         childFactory = { bottomSheetFile, _ -> bottomSheetFile }
     )
 
-    private val selectionViewModel = instanceKeeper.getOrCreate(path.toString()) {
+    private val selectionViewModel = instanceKeeper.getOrCreate("selectionViewModel_$path") {
```
LionZXY, line 60: Do you think it's necessary? As far as I remember the viewmodel class name is already used for the unique key. Can you explain why it is necessary?

LionZXY (approved):

CHANGELOG.md
```diff
@@ -70,6 +70,7 @@ Attention: don't forget to add the flag for F-Droid before release
 - [FIX] Add deeplink fallback for flipper scheme uri
 - [FIX] Change description of remotes library card
 - [FIX] Fix bytes race in new api
+- [FIX] New file manager uploading progress
```
LionZXY, line 73: Also, please, change changelog

components/filemngr/listing/impl/src/commonMain/kotlin/com/flipperdevices/filemanager/listing/impl/api/FilesDecomposeComponentImpl.kt
makeevrserg: `filesViewModel` and `selectionViewModel` previously were using same key "$path" before it crashed. The crash was becaseu intanceKeeper got class by key "$path" and the class was `filesViewModel`. But should be `selectionViewModel`. So the instance keeper compare only key, not class itself. Due to this implementation classes with "same key" needs to add unique prefix.

The reason why this is required inside InstanceKeeper impl:
```kotlin
check(key !in map) { "Another instance is already associated with the key: $key" }
```

LionZXY (approved):

components/filemngr/listing/impl/src/commonMain/kotlin/com/flipperdevices/filemanager/listing/impl/api/FilesDecomposeComponentImpl.kt
LionZXY: > So the instance keeper compare only key, not class itself.

That's weird, I've been re-doing that part - let's discuss in a PM after your vacation

PR #992, opened 2024-11-25
Single tap infrared

makeevrserg:
**Background**

When pressing infrared buttons anywhere - the signal will be send not once. FW added new request which allows one-time infrared signal send

**Changes**

- Change EmulateApi with new one-time send

**Test plan**

- Build firmware from https://github.com/flipperdevices/flipperzero-firmware/pull/4000
- Open flipper app
- Open some infrared key
- Press key once and see emulating on flipper now shorter
- Repeat this step for Infrared Remote Controls


components/keyemulate/impl/src/main/java/com/flipperdevices/keyemulate/helpers/StartEmulateHelper.kt
```diff
@@ -125,7 +127,17 @@ class StartEmulateHelperImpl @Inject constructor(
         val appButtonPressResponse = serviceApi.requestApi.request(
             flowOf(
                 main {
-                    appButtonPressRequest = getAppButtonPressRequest(config, isIndexEmulateSupport)
+                    if (config.isPressRelease) {
+                        appButtonPressReleaseRequest = getAppButtonPressReleaseRequest(
+                            config,
```
makeevrserg, line 132: Need to add version check for old flippers where this feature was not yet in firmware

makeevrserg: Added version checks

components/bridge/api/src/main/java/com/flipperdevices/bridge/api/utils/Constants.kt
```diff
@@ -33,6 +33,10 @@ object Constants {
         majorVersion = 0,
         minorVersion = 21
     )
+    val API_SUPPORTED_INFRARED_PRESS_RELEASE = SemVer(
+        majorVersion = 0,
+        minorVersion = 9999999 // TODO
```
makeevrserg, line 38: Fix this after firmware release

LionZXY (approved):

PR #1000, opened 2024-12-10
Connection service for new API

makeevrserg:
**Background**

Device connection service for new API

**Changes**

- Add new connection service
- Start it in application

**Test plan**

- Install flipper from branch, open int
- See it now bottom bar display connection state and synchronization

### DIFF
--- a/components/bridge/connection/service/api/src/commonMain/kotlin/com/flipperdevices/bridge/connection/service/api/FConnectionService.kt
+++ b/components/bridge/connection/service/api/src/commonMain/kotlin/com/flipperdevices/bridge/connection/service/api/FConnectionService.kt
@@ -1,0 +1,22 @@
+package com.flipperdevices.bridge.connection.service.api
+
+/**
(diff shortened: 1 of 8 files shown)


components/bridge/connection/service/api/src/commonMain/kotlin/com/flipperdevices/bridge/connection/service/api/FConnectionService.kt
```diff
@@ -0,0 +1,22 @@
...
+interface FConnectionService {
+    /**
+     * Should be called once on application start
+     */
+    fun onApplicationInit()
+
+    /**
+     * Reconnect to last known device after being force disconnected
+     */
+    fun forceReconnect()
```
makeevrserg, line 15: `forceReconnect` and `disconnect` may be changed in future to something else. Don't have enough knowledge of overall connectivity thing

LionZXY (approved):

PR #1008, opened 2024-12-25
Migrate update to new BLE api

makeevrserg:
**Background**

Migration of all update modules into new api

**Changes**

- Added new feature for update
- Migrated every update submodules into new api

**Test plan**

- Open flipper android app and run update
- See flipper is update correctly with all it's flow (image changed, rebooted, location sub-ghz working)
- See all tests previously located is passed

### TODO

- Device reconnect after it being shutdown(we don't get success message on mobile)(the problem is on another layer. Not actually related to feature)

### DIFF
--- a/components/updater/card/src/main/java/com/flipperdevices/updater/card/viewmodel/UpdateRequestViewModel.kt
+++ b/components/updater/card/src/main/java/com/flipperdevices/updater/card/viewmodel/UpdateRequestViewModel.kt
@@ -46,11 +46,13 @@ class UpdateRequestViewModel @Inject constructor(
         .map { gattInfo ->
             if (gattInfo == null) {
                 BatteryState.Unknown
-
             } else {
-                BatteryState.Ready(gattInfo.isCharging, gattInfo.batteryLevel ?: 0f)
+                gattInfo.batteryLevel?.let { batteryLevel ->
+                    BatteryState.Ready(gattInfo.isCharging, batteryLevel)
+                } ?: BatteryState.Unknown
             }
-        }.stateIn(viewModelScope, SharingStarted.Eagerly, BatteryState.Unknown)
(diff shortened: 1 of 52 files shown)


components/updater/card/src/main/java/com/flipperdevices/updater/card/viewmodel/UpdateRequestViewModel.kt
```diff
@@ -30,17 +36,29 @@ private const val EXT_UPDATER_FILE = "tgz"
...
+        .flatMapLatest { status -> status?.featureApi?.getGattInfoFlow() ?: flowOf(null) }
+        .map { gattInfo ->
+            if (gattInfo == null) {
+                BatteryState.Unknown
+            } else {
+                gattInfo.batteryLevel?.let { batteryLevel ->
+                    BatteryState.Ready(gattInfo.isCharging, batteryLevel)
+                } ?: BatteryState.Unknown
+            }
+        }.stateIn(viewModelScope, SharingStarted.Lazily, BatteryState.Unknown)
```
makeevrserg, line 54: Need to be lazily because we can't properly mock FGattInfoFeatureApi without it

components/updater/card/src/test/kotlin/com/flipperdevices/updater/card/viewmodel/UpdateRequestViewModelTest.kt
```diff
@@ -124,6 +136,9 @@ class UpdateRequestViewModelTest {
     @Test
     fun `Open update request with sync`() = runTest {
         synchronizationState.emit(SynchronizationState.InProgress.Default(0f))
+        viewModel.getBatteryState()
```
makeevrserg, line 139: It's required to call batteryState initialization. It's lazy initialization

LionZXY (approved):

PR #1019, opened 2025-01-22
Fix infinite disconnect

makeevrserg:
**Background**

This PR fixes bug with infinite disconnection

**Changes**
- Fix scope cancellation for device
- Add custom scope for device connection which is cancelled on disconnect
- Fix reconnect after flipper reboot
- Fix some detekt issues
- Fix reconnect after device is powered-off for a long time
- Fix update

**Test plan**
- Open app
- Wait for connection
- Disconnect
- Connect again
- Repeat a few times
- See everything is ok

### DIFF
--- a/components/bridge/connection/service/impl/src/commonMain/kotlin/com/flipperdevices/bridge/connection/service/impl/FConnectionServiceImpl.kt
+++ b/components/bridge/connection/service/impl/src/commonMain/kotlin/com/flipperdevices/bridge/connection/service/impl/FConnectionServiceImpl.kt
@@ -40,7 +40,7 @@ class FConnectionServiceImpl @Inject constructor(
         return orchestrator.getState()
             .onEach {
                 if (it !is FDeviceConnectStatus.Disconnected) return@onEach
-                when(it.reason) {
+                when (it.reason) {
                     DisconnectStatus.NOT_INITIALIZED -> return@onEach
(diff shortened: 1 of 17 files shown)


components/bridge/connection/service/impl/src/commonMain/kotlin/com/flipperdevices/bridge/connection/service/impl/FConnectionServiceImpl.kt
```diff
@@ -32,26 +36,50 @@ class FConnectionServiceImpl @Inject constructor(
...
+        return orchestrator.getState()
+            .onEach {
+                if (it !is FDeviceConnectStatus.Disconnected) return@onEach
+                when (it.reason) {
+                    DisconnectStatus.NOT_INITIALIZED -> return@onEach
+                    DisconnectStatus.REPORTED_BY_TRANSPORT -> Unit
+                    DisconnectStatus.ERROR_UNKNOWN -> Unit
+                }
+                if (isForceDisconnected.first()) return@onEach
+                val currentDevice = fDevicePersistedStorage.getCurrentDevice()
```
LionZXY, line 49: You can use `state.device` here

makeevrserg: Fixed!

LionZXY (approved):

PR #1021, opened 2025-01-24
Add usb connection for sample

LionZXY:
**Background**

This is first step to add USB connection to flipper mobile app

**Changes**

- Add USB android library and base implementation of search and USB connection

image

**Test plan**

Try to connect flipper by usb and use connection sample


makeevrserg (dismissed):

kamalg777:
Not working

Issue #1032, opened 2025-06-07
Multi-Flipper support


lucidnx:
Hello! Can you add support for multiple flipper devices? currently I need to remove flipper from app, connect to another and wait for sync to do it's job..
I expect:
1. connect menu where I can select device
2. sync should work for each device and synced data will persist even if you connect to another flipper.

Thanks! :)

PR #1038, opened 2025-07-14
display mac address in device list

0xDE57:
add mac address field to device list

**Background**

More verbosity for user to see device info.

**Changes**

Adds a new column with mac address underneath device name.

### DIFF
--- a/components/firstpair/impl/src/main/java/com/flipperdevices/firstpair/impl/composable/searching/ComposableSearchItem.kt
+++ b/components/firstpair/impl/src/main/java/com/flipperdevices/firstpair/impl/composable/searching/ComposableSearchItem.kt
@@ -34,6 +34,7 @@ import com.flipperdevices.firstpair.impl.R
 @Suppress("LongMethod")
 fun ComposableSearchItem(
     text: String,
+    underText: String,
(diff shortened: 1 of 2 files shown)


LionZXY:
To me it looks like an over-complication of the interface. Do you think having a MAC address will change anything for the user?
I think it will make the interface worse, not better

0xDE57:
I like the extra verbosity personally, and I don't think it's over-complicated because the target audience of the flipper is generally more technical users. 

Unless by complex you mean busy, yes it does make the screen / list busier. Maybe it could be adjusted for better aesthetics? eg: mac address a bit smaller font or something

It maybe changes things for user's with multiple flippers, for them to see more detail on each device? For example if this feature ever gets added: #1032

IDk. It's your project, your choice. ¯\_(ツ)_/¯

LionZXY:
It seems to me that **extra verbose information** on the page makes the **interface more complex**. 
It makes the **application less user friendly**.

I don't see any reason (**now**) to show the mac address with the condition of a unique Flipper name, but I'd be happy to revise this issue if such a use-case appears

0xDE57:
Okey dokes. :+1:

